PR #4, opened 2019-07-08
build: add build script

ajeddeloh:
cc @arithx


build
```diff
@@ -0,0 +1,18 @@
...
+
+NAME=fcct
+export GOFLAGS=-mod=readonly
+
+eval $(go env)
+if [ -z ${BIN_PATH+a} ]; then
+	BIN_PATH=${PWD}/bin/${GOARCH}
+fi
+
+export CGO_ENABLED=1
```
arithx: minor: let's group all the exports together at the top

ajeddeloh: fixed.

PR #5, opened 2019-07-09
Add release tooling

ajeddeloh:
Add release tagging/building scripts as well as a checklist.

### DIFF
--- a/build_releases
+++ b/build_releases
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+export GO111MODULE=on
+export GOFLAGS=-mod=readonly
+export CGO_ENABLED=0
+
+eval $(go env)
+if [ -z ${BIN_PATH+a} ]; then
+	export BIN_PATH=${PWD}/bin/releases/
+fi
+
+build_release() {
+    export NAME="fcct-${1}"
+    echo "building ${NAME}"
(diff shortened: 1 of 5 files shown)


build_releases
```diff
@@ -0,0 +1,34 @@
...
+
+build_release() {
+    export NAME="fcct-${1}"
+    echo "building ${NAME}"
+    go build -o ${BIN_PATH}/${NAME} internal/main.go
+}
+
+export GOOS=linux
+export GOARCH=amd64
+build_release x86_64-unknown-linux-gnu
```
jlebon: Minor/optional: could just make `build_release` take the OS and arch as paramters.

build_releases
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -eu
```
jlebon: The bash scripter in me is itching to add a `o pipefail` to that.

tag_release.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+[ $# == 2 ] || { echo "usage: $0 version commit" && exit 1; }
```
jlebon: `<version> <commit>` ?

build_releases
ajeddeloh: eh, I'll leave it off until it gets hairy. This matches the Ignition one too.

ajeddeloh:
fixed.

PR #10, opened 2019-07-17
main: Fix output file handling

sgreene570:
This will let you actually write an ignition file with the `--output` flag.
Also added a newline print to the error printing function.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -48,31 +48,31 @@ func main() {
 		var err error
 		infile, err = os.Open(input)
 		if err != nil {
-			fail("failed to open %s: %v", input, err)
+			fail("failed to open %s: %v\n", input, err)
 		}
 		defer infile.Close()
 	}
 
 	dataIn, err := ioutil.ReadAll(infile)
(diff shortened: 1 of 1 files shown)


internal/main.go
```diff
@@ -26,6 +26,7 @@ import (
 
 func fail(format string, args ...interface{}) {
 	fmt.Fprintf(os.Stderr, format, args...)
+	fmt.Println()
```
ajeddeloh: this should be added to the fmt string where fail() is called instead.

ashcrow: IE:

```golang
fmt.Fprintf(os.Stderr, fmt.Sprintf("%s\n", format), args...)
```

or added to each call of `fail()`'s format.

ajeddeloh: added to each call of fail() I think.

sgreene570: will do

sgreene570:
should be good to go now.

sgreene570:
Closes #12

PR #21, opened 2019-07-19
Docs: Update gpg verify instructions

### LINKED ISSUE #13: gpg verify instructions don't work
Getting started doc says:
```
Download the latest version of fcct and the detached signature from the releases page. Verify it with gpg:
gpg --verify <fcct binary> <detached sig>
```
However:
```
$ gpg --verify fcct-x86_64-unknown-linux-gnu fcct-x86_64-unknown-linux-gnu.asc 
gpg: no valid OpenPGP data found.
gpg: the signature could not be verified.
Please remember that the signature file (.sig or .asc)
should be the first file given on the command line.
```
So it seems that those two files should be swapped:
```
$ gpg --verify fcct-x86_64-unknown-linux-gnu.asc fcct-x86_64-unknown-linux-gnu
gpg: Signature made Thu 11 Jul 2019 12:46:56 AM CEST
gpg:                using RSA key 9CEB8FE6B4F1E9E752F61C82CDDE268EBB729EC7
gpg: Can't check signature: No public key
```
That works better, but complains about no public key. I'm not very fluent with gpg, so I might be doing something wrong.

sgreene570:
Closes #13

Add quick instructions on how to import CoreOS application public key (just in case)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -11,6 +11,7 @@ Download the latest version of `fcct` and the detached signature from the [relea
 ```
 gpg --verify <detached sig> <fcct binary>
 ```
+You may need to download the [CoreOS Application Signing Key](http://coreos.com/security/app-signing-key/) and import it with `gpg --import <key>` if you have not already done so.
 
 New releases of `fcct` are backwards compatible with old releases unless otherwise noted.
 


docs/getting-started.md
```diff
@@ -11,6 +11,7 @@ Download the latest version of `fcct` and the detached signature from the [relea
 ```
 gpg --verify <detached sig> <fcct binary>
 ```
+(You may need to download the [CoreOS Application Signing Key](http://coreos.com/security/app-signing-key/) and import it with `gpg --import <key>` if you have not already done so)
```
arithx: Minor: drop the parens around the whole line and add a trailing period.

sgreene570: Thanks, fixed.

PR #22, opened 2019-07-19
Dockerfile: add dockerfile for releases

ajeddeloh:
Add a dockerfile for building release containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:latest AS builder
+RUN mkdir /fcct
+COPY . /fcct
+WORKDIR /fcct
+RUN ./build_releases
+
+FROM scratch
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:latest AS builder
+RUN mkdir /fcct
+COPY . /fcct
+WORKDIR /fcct
+RUN ./build_releases
+
+FROM scratch
+COPY --from=builder /fcct/bin/releases/fcct-x86_64-unknown-linux-gnu /bin/fcct
```
lucab: I usually go for `/usr/local/bin`, but the result is the same.

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+ENTRYPOINT ["/bin/fcct"]
+
```
ashcrow: nit: Extra new line

PR #27, opened 2019-07-22
test: add checking for examples in docs, travis

ajeddeloh:
Extract and try all the examples in the docs as part of the test process
to ensure we don't include invalid examples.

Fixes https://github.com/coreos/fcct/issues/24
Fixes https://github.com/coreos/fcct/issues/23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+go:
+  - "1.12.x"
+
+install: echo nop
+
+script:
+  - ./test
--- a/test
+++ b/test
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -euo pipefail
 
 export GO111MODULE=on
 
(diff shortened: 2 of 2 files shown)


test
```diff
@@ -8,5 +8,21 @@ res=$(gofmt -d .)
...
+shopt -s nullglob
+echo "Checking docs"
+for doc in docs/*md
+do
+	echo "Checking $doc"
+	# split each doc into a bunch of tmpfiles then run ct on them
+	sed -n '/^```yaml fedora-coreos-config/,/^```/ p' < ${doc} | csplit - '/```yaml fedora-coreos-config/' '{*}' -z --prefix tmpconfig -q
+	for i in tmpconfig*
+	do
+		cat "$i" | tail -n +2 | head -n -1 | go run internal/main.go > /dev/null || cat -n "$i"
```
arithx: Should probably do a `--strict` run. Our docs shouldn't have any warnings.

test
```diff
@@ -8,5 +8,21 @@ res=$(gofmt -d .)
...
+		rm "$i"
+	done
+done
+
+echo ok
```
arithx: This ends up printing regardless of if the docs have errors, is that intentional?

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+go:
+  - "1.12.x"
+
+install: echo nop
```
ashcrow: Is this needed?

ajeddeloh: Yes. Travis has default behavior to instead go deps that needs an env var set to use go modules correctly. I'd set the envvar but we end up doing what it is doing anyway in `./test`. So this both avoids the problem and makes the test run a little faster. I'll add a comment.

ajeddeloh:
Fixed.

test
```diff
@@ -8,5 +8,21 @@ res=$(gofmt -d .)
...
+echo "Checking docs"
+for doc in docs/*md
+do
+	echo "Checking $doc"
+	# split each doc into a bunch of tmpfiles then run ct on them
+	sed -n '/^```yaml fedora-coreos-config/,/^```/ p' < ${doc} | csplit - '/```yaml fedora-coreos-config/' '{*}' -z --prefix tmpconfig -q
+	for i in tmpconfig*
+	do
+		cat "$i" | tail -n +2 | head -n -1 | go run internal/main.go > /dev/null || cat -n "$i"
+		rm "$i"
```
arithx: These files are not deleted if there's an error, should probably add a trap to delete them on exit and cancel it out after the loop.

ajeddeloh: I think GH is having a problem. Look at my branch this PR is based on and it's fixed there.

ajeddeloh:
Updated to use trap, GH seems to be working now.

PR #33, opened 2019-07-25
docs/getting-started: add fcct container docs

ajeddeloh:
We publish fcct containers on quay. Document that and add instructions.

Fixes https://github.com/coreos/fcct/issues/19

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -6,6 +6,10 @@ Fedora CoreOS Configs are YAML files conforming to `fcct`'s schema. For more inf
 
 ### Getting FCCT
 
+`fcct` can be downloaded as a standalone binary or run as a container with docker or podman.
+
+#### Standalone binary
+
 Download the latest version of `fcct` and the detached signature from the [releases page](https://github.com/coreos/fcct/releases). Verify it with gpg:
 
 ```
@@ -14,6 +18,21 @@ gpg --verify <detached sig> <fcct binary>
 You may need to download the [CoreOS Application Signing Key](http://coreos.com/security/app-signing-key/) and import it with `gpg --import <key>` if you have not already done so.
 
 New releases of `fcct` are backwards compatible with old releases unless otherwise noted.
+
+#### Container
+
+This example uses podman, but docker can also be used. Substitute v0.2.0 with the desired version. Note that the `latest` tag corresponds with `master` and not with the latest release.
+
+```bash
+# Pull the desired version
+podman pull quay.io/coreos/fcct:v0.2.0
+
+# Run fcct using standard in and standard out
+podman run -it --rm quay.io/coreos/fcct:v0.2.0 -pretty -strict < your_config.fcc > transpiled_config.ign
+
+# Run fcct using files.
+podman run --rm -v your_config.fcc:/config.fcc:z quay.io/coreos/fcct:v0.2.0 -pretty -strict -input /config.fcc > transpiled_config.ign
+```
 
 ### Writing and using Fedora CoreOS Configs
 


docs/getting-started.md
```diff
@@ -15,6 +19,21 @@ You may need to download the [CoreOS Application Signing Key](http://coreos.com/
...
+
+```bash
+# Pull the desired version
+podman pull quay.io/coreos/fcct:v0.2.0
+
+# Run fcct using standard in and standard out
+podman run -it --rm fcct --pretty --strict < your_config.fcc > transpiled_config.ign
+
+# Run fcct using files. Note that the directory containing the files must be mounted into the container
+podman run --rm -v path_to_config_dir:/srv fcct -input /srv/your_config.fcc -output /srv/transpiled_config.ign --pretty --strict
```
arithx: While it's still valid we should probably not mix single and double dashes on the parameters.

ajeddeloh: Fixed. Also added a note about the `latest` tag

docs/getting-started.md
```diff
@@ -15,6 +19,21 @@ You may need to download the [CoreOS Application Signing Key](http://coreos.com/
 
 New releases of `fcct` are backwards compatible with old releases unless otherwise noted.
 
+#### Container
+
+This example uses, but docker can also be used. Substitute v0.2.0 with the desired version. Note that the `latest` tag corresponds with `master` and not with the latest release.
```
lucab: "This example uses **podman**, but ..."

ajeddeloh: Good catch, fixed.

docs/getting-started.md
```diff
@@ -15,6 +19,21 @@ You may need to download the [CoreOS Application Signing Key](http://coreos.com/
...
+#### Container
+
+This example uses podman, but docker can also be used. Substitute v0.2.0 with the desired version. Note that the `latest` tag corresponds with `master` and not with the latest release.
+
+```bash
+# Pull the desired version
+podman pull quay.io/coreos/fcct:v0.2.0
+
+# Run fcct using standard in and standard out
+podman run -it --rm fcct -pretty -strict < your_config.fcc > transpiled_config.ign
```
jlebon: Hmm, doesn't this need to be the full pullspec again? So e.g. `podman run -it --rm quay.io/coreos/fcct:v0.2.0 ...`?

ajeddeloh: Not sure, but having it doesn't hurt and is definitely more clear. Fixed

docs/getting-started.md
```diff
@@ -15,6 +19,21 @@ You may need to download the [CoreOS Application Signing Key](http://coreos.com/
...
+
+```bash
+# Pull the desired version
+podman pull quay.io/coreos/fcct:v0.2.0
+
+# Run fcct using standard in and standard out
+podman run -it --rm fcct:v0.2.0 -pretty -strict < your_config.fcc > transpiled_config.ign
+
+# Run fcct using files. Note that the directory containing the files must be mounted into the container
+podman run --rm -v path_to_config_dir:/srv fcct:v0.2.0 -input /srv/your_config.fcc -output /srv/transpiled_config.ign -pretty -strict
```
jlebon: Hmm, were you able to test this? On Fedora/RHEL, I'd expect SELinux permission issues.

Another approach is to bind-mount the file directly with `:z` so it gets relabeled. So e.g.:

```
podman run --rm -v path_to_fcc_file:/config.fcc:z fcct:v0.2.0 -input /config.fcc > output.ign
```

The advantage of this is that you don't have to bind-mount and relabel a whole directory, which is a common gotcha for a lot of people. E.g. doing `-v /home/bob:/srv:z` will wreck your system.

(We could even default the container to use `-input /config.fcc` to make this easier.)

ajeddeloh: I did test this but on my laptop which doesn't have SELinux enabled.

I don't like the idea of `-input /config.fcc` being default because it's less intuitive to not have the default be stdin/stdout. It's also different than the default when not running it in a container.

jlebon: > I don't like the idea of -input /config.fcc being default because it's less intuitive to not have the default be stdin/stdout

Yeah, makes sense. WDYT about the rest of the proposal though? (I.e. bind-mounting the file directly so podman only relabels that bit.)

Alternatively, we could just add `:z` to what you currently have, but mention that the whole directory will be relabeled if it's not already the right label?

ajeddeloh: I like that, just updated to that.

docs/getting-started.md
lucab: What's the final direction on this? Are you still planning to change this to use the fully-specified image identifier?

ajeddeloh: Do you mean with regard to latest being master or referring to the container as `fcct:version` vs `quay.io/coreos/fcct:version`?

lucab: The latter (sorry, I misplaced the comment above, it was a followup to @jlebon thread).

ajeddeloh:
switched to full pullspecs

PR #37, opened 2019-07-29
[docs] passwords are hashed not encrypted

stbischof:
password_hashes should be hashes not encrypted


lucab:
LGTM, thanks for the doc-fix!

PR #40, opened 2019-08-15
config/v1_0/fcos.go: don't unconditionally print newline

jlebon:
Use `fmt.Print()` instead of `fmt.Println()` to print the report. The
`String()` method already adds a newline at the end of each report
entry.

My main motivation for doing this though is a super tiny polish item:
even `fcct` invocations fed perfectly valid YAML would still print an
empty line:

```
$ fcct -input basic.yaml -output basic.json

$
```

### DIFF
--- a/config/v1_0/fcos.go
+++ b/config/v1_0/fcos.go
@@ -83,7 +83,7 @@ func TranslateBytes(input []byte, options common.TranslateOptions) ([]byte, erro
 	second := validate.Validate(final, "json")
 	second.Correlate(translatedTree)
 	r.Merge(second)
-	fmt.Println(r.String())
+	fmt.Print(r.String())
 
 	if r.IsFatal() {
 		return nil, ErrInvalidConfig


config/v1_0/fcos.go
```diff
@@ -83,7 +83,7 @@ func TranslateBytes(input []byte, options common.TranslateOptions) ([]byte, erro
 	second := validate.Validate(final, "json")
 	second.Correlate(translatedTree)
 	r.Merge(second)
-	fmt.Println(r.String())
```
lucab: You hit this specific one, however we use this idiom pretty much everywhere.
I think it's more effective if we stuck a `strings.TrimSpace()` either in the `String` method or in `Correlate`/`Merge`.

jlebon: > I think it's more effective if we stuck a strings.TrimSpace() either in the String method or in Correlate/Merge

Hmm, not sure I follow. The unconditional newline comes from the `Println()` here, not `r.String()`.

lucab: Ah, my understanding was that you got two newlines because "The `String()` method already adds a newline at the end of each report entry."

lucab:
@ajeddeloh it looks like my previous review on this one was a bit off. As you have better insights into the flow here, mind having a look?

ajeddeloh:
Sorry, forgot about this, really that package shouldn't be printing at all. Thoughts on https://github.com/coreos/fcct/pull/49 ?

PR #50, opened 2019-10-08
Fix line and column handling and propagation

ajeddeloh:
In this PR:
 - Fork Ignition's translation code - we need to do more complex things than it ever did
 - Update the translation code to also emit a list of (src, dest) paths. For example you might have (`$yaml.storage.files.4.contents.inline`, `$json.storage.files.4.contents.source`) meaning the 5th element of the files in the generating config got it's source field from the corresponding inline field in the fcc. 
 - Use that list of translations to convert a report with ignition paths to fcc paths, then correlate that to the actual source for line/col information.
 - Add a bunch of helpers to make writing translators for that easier
 - Update tests for handle all this and test the new functionality

### DIFF
--- a/base/v0_1/translate.go
+++ b/base/v0_1/translate.go
@@ -17,75 +17,85 @@ package v0_1
 import (
 	"net/url"
 
-	"github.com/coreos/ignition/v2/config/translate"
+	"github.com/coreos/fcct/translate"
+
 	"github.com/coreos/ignition/v2/config/v3_0/types"
+	"github.com/coreos/vcontext/path"
 	"github.com/vincent-petithory/dataurl"
 )
 
-func (c Config) ToIgn3_0() (types.Config, error) {
+// ToIgn3_0 translates the config to an Ignition config. It also returns the set of translations
+// it did so paths in the resultant config can be tracked back to their source in the source config.
+func (c Config) ToIgn3_0() (types.Config, translate.TranslationSet, error) {
 	ret := types.Config{}
-	tr := translate.NewTranslator()
+	tr := translate.NewTranslator("yaml", "json")
 	tr.AddCustomTranslator(translateIgnition)
 	tr.AddCustomTranslator(translateFile)
 	tr.AddCustomTranslator(translateDirectory)
 	tr.AddCustomTranslator(translateLink)
-	tr.Translate(&c, &ret)
-	return ret, nil
+	translations := tr.Translate(&c, &ret)
+	return ret, translations, nil
 }
 
-func translateIgnition(from Ignition) (to types.Ignition) {
-	tr := translate.NewTranslator()
+func translateIgnition(from Ignition) (to types.Ignition, tm translate.TranslationSet) {
+	tr := translate.NewTranslator("yaml", "json")
 	to.Version = types.MaxVersion.String()
-	tr.Translate(&from.Config, &to.Config)
-	tr.Translate(&from.Security, &to.Security)
-	tr.Translate(&from.Timeouts, &to.Timeouts)
+	tm = tr.Translate(&from.Config, &to.Config).Prefix("config")
+	tm.MergeP("security", tr.Translate(&from.Security, &to.Security))
+	tm.MergeP("timeouts", tr.Translate(&from.Timeouts, &to.Timeouts))
 	return
 }
 
-func translateFile(from File) (to types.File) {
-	tr := translate.NewTranslator()
+func translateFile(from File) (to types.File, tm translate.TranslationSet) {
+	tr := translate.NewTranslator("yaml", "json")
 	tr.AddCustomTranslator(translateFileContents)
-	tr.Translate(&from.Group, &to.Group)
-	tr.Translate(&from.User, &to.User)
-	tr.Translate(&from.Append, &to.Append)
-	tr.Translate(&from.Contents, &to.Contents)
+	tm = tr.Translate(&from.Group, &to.Group).Prefix("group")
+	tm.MergeP("user", tr.Translate(&from.User, &to.User))
+	tm.MergeP("append", tr.Translate(&from.Append, &to.Append))
+	tm.MergeP("contents", tr.Translate(&from.Contents, &to.Contents))
 	to.Overwrite = from.Overwrite
 	to.Path = from.Path
 	to.Mode = from.Mode
+	tm.AddIdentity("overwrite", "path", "mode")
 	return
 }
 
-func translateFileContents(from FileContents) (to types.FileContents) {
+func translateFileContents(from FileContents) (to types.FileContents, tm translate.TranslationSet) {
+	tr := translate.NewTranslator("yaml", "json")
+	tm = tr.Translate(&from.Verification, &to.Verification).Prefix("verification")
 	to.Source = from.Source
 	to.Compression = from.Compression
-	to.Verification.Hash = from.Verification.Hash
+	tm.AddIdentity("source", "compression")
 	if from.Inline != nil {
 		src := (&url.URL{
 			Scheme: "data",
 			Opaque: "," + dataurl.EscapeString(*from.Inline),
 		}).String()
 		to.Source = &src
+		tm.AddTranslation(path.New("yaml", "inline"), path.New("json", "source"))
 	}
 	return
 }
 
-func translateDirectory(from Directory) (to types.Directory) {
-	tr := translate.NewTranslator()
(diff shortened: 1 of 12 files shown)


base/v0_1/translate.go
```diff
@@ -17,75 +17,83 @@ package v0_1
...
-	"github.com/coreos/ignition/v2/config/translate"
+	"github.com/coreos/fcct/translate"
+
 	"github.com/coreos/ignition/v2/config/v3_0/types"
+	"github.com/coreos/vcontext/path"
 	"github.com/vincent-petithory/dataurl"
 )
 
-func (c Config) ToIgn3_0() (types.Config, error) {
+func (c Config) ToIgn3_0() (types.Config, translate.TranslationSet, error) {
```
ashcrow: nit: The name is pretty obvious, but I'm a stickler for function documentation ... especially when it's public functions.

config/common/common.go
```diff
@@ -55,41 +64,34 @@ func Marshal(from interface{}, pretty bool) ([]byte, error) {
...
-		for _, v := range n.Children {
-			s.Children = append(s.Children, ToCamelCase(v))
+// snake converts from camelCase (not CamelCase) to snake_case
+func snake(in string) string {
+	return strings.ToLower(snakeRe.ReplaceAllString(in, "_$1"))
+}
+
+// ConvertReport takes a report from a camelCase json document and a set of translations rules,
+// applies those rules and converts all camelCase to snake_case.
+func ConvertReport(r *report.Report, ts translate.TranslationSet) {
```
ashcrow: very minor nit: `ConvertReport` can be read 2 ways:

- Produce a report on the converting
- Use a report and convert

Consider `ConvertFromReport`.

However, the doc above does a great job explaining what the function does :+1:.

translate/translate.go
```diff
@@ -0,0 +1,313 @@
...
+// of the destination path to the translation struct. The map is purely an optimization to allow fast lookups. Ideally the
+// map would just be from the destination path.ContextPath to the source path.ContextPath, but ContextPath contains a slice
+// which are not comparable and thus cannot be used as keys in maps.
+type TranslationSet struct {
+	FromTag string
+	ToTag   string
+	Set     map[string]Translation
+}
+
+func (ts TranslationSet) AddTranslation(from, to path.ContextPath) {
```
ashcrow: nit: Public doc

base/v0_1/translate_test.go
```diff
@@ -18,21 +18,49 @@ import (
...
 
 // Most of this is covered by the Ignition translator generic tests, so just test the custom bits
 
+// verifyTranslations ensures all the translations are identity, unless they match a listed one
+// it returns the offending translation if there is one
+func verifyTranslations(set translate.TranslationSet, exceptions ...translate.Translation) *translate.Translation {
+	exceptionSet := translate.TranslationSet{
+		FromTag: set.FromTag,
+		ToTag:   set.ToTag,
+		Set:     map[string]translate.Translation{},
```
ashcrow: I see `map[string]Translation` used a number of times in the code. Consider making it it's own type.

translate/translate.go
```diff
@@ -0,0 +1,313 @@
...
+	case util.IsPrimitive(k1):
+		return true
+	case util.IsInvalidInConfig(k1):
+		panic(fmt.Sprintf("Encountered invalid kind %s in config. This is a bug, please file a report", k1))
+	case k1 == reflect.Ptr || k1 == reflect.Slice:
+		return t.translatable(t1.Elem(), t2.Elem()) || t.hasTranslator(t1.Elem(), t2.Elem())
+	case k1 == reflect.Struct:
+		return t.translatableStruct(t1, t2)
+	default:
+		panic(fmt.Sprintf("Encountered unknown kind %s in config. This is a bug, please file a report", k1))
```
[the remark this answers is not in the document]
arithx: Personally I feel it's fine to die immediately, it's an extreme corner case.

translate/translate.go
```diff
@@ -79,42 +85,66 @@ func couldBeValidTranslator(t reflect.Type) bool {
...
 		return false
 	}
-	if t.NumIn() != 1 || t.NumOut() != 1 {
+	if t.NumIn() != 1 || t.NumOut() != 2 {
 		return false
 	}
-	if util.IsInvalidInConfig(t.In(0).Kind()) || util.IsInvalidInConfig(t.Out(0).Kind()) {
+	if util.IsInvalidInConfig(t.In(0).Kind()) ||
+		util.IsInvalidInConfig(t.Out(0).Kind()) ||
+		t.Out(1) != translationsType {
```
arithx: _sigh,_ `gofmt--`

translate/translate.go
```diff
@@ -79,42 +85,66 @@ func couldBeValidTranslator(t reflect.Type) bool {
...
 		return false
 	}
 	return true
 }
 
+// fieldName returns the name uses when (un)marshalling a field. t should be a reflect.Value of a struct,
+// index is the field index, and tag is the struct tag used when (un)marshalling (e.g. "json" or "yaml")
+func fieldName(t reflect.Value, index int, tag string) string {
```
arithx: Should this also do an initial assert that `t.Kind() == reflect.Struct`?

translate/translate.go
```diff
@@ -79,42 +85,66 @@ func couldBeValidTranslator(t reflect.Type) bool {
...
 		vTo.Set(reflect.MakeSlice(vTo.Type(), vFrom.Len(), vFrom.Len()))
 		for i := 0; i < vFrom.Len(); i++ {
-			t.translate(vFrom.Index(i), vTo.Index(i))
+			t.translate(vFrom.Index(i), vTo.Index(i), fromPath.Append(i), toPath.Append(i))
 		}
 	case k == reflect.Struct:
 		for i := 0; i < vFrom.NumField(); i++ {
-			t.translate(vFrom.Field(i), vTo.FieldByName(vFrom.Type().Field(i).Name))
+			fieldGoName := vFrom.Type().Field(i).Name
+			toStructField, _ := vTo.Type().FieldByName(fieldGoName)
```
arithx: Might be better to not throw away the `bool` return of `FieldByName` and `panic` with a more reasonable message than allowing the follow-up line to panic at index out of bounds / null object.

translate/translate_test.go
```diff
@@ -0,0 +1,205 @@
...
+// limitations under the License.
+
+package translate
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/coreos/ignition/v2/config/translate/tests/pkga"
```
arithx: Should be using the `fcct` paths rather than the `Ignition` paths.

config/common/common.go
ajeddeloh: I don't like ConvertFromReport since that implies the result is not a report. 

What about `TranslateReportPaths()`?

translate/translate.go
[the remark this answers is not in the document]
ajeddeloh: This is very intentional. The check for translatablity operates mainly on types and thus isn't dependent on the contents of the config. If this is hit, it's hit immediately when translating regardless of if that part of a config is filled out or not. This means a config structure with a bad kind will _always_ panic. Basically the translator is unusable and won't work at all, wont pass ci, etc. It's like a compile time check at runtime.

base/v0_1/translate_test.go
ajeddeloh: That's basically what TranslationSet is. In practice users don't initialize TranslationSets themselves much since Translate() does that. It's basically just tests where we need to have a bunch a literals to compare against, so I'm inclined to leave it as is.

translate/translate.go
ajeddeloh: And do what? We could make it an error but that would only come up due to programming errors and would be propagated all over the place. We could make it panic that's what it already does.

translate/translate.go
ajeddeloh: Fair, it should never be hit since the translatable check should handle that, but I'm all for better panic messages.

translate/translate_test.go
ajeddeloh: This gets fixed in the next commit. I wanted to copy-over to be verbatim so it's easier to see what changed exactly.

ajeddeloh:
Fixed most of the comments, responded to others.

translate/translate.go
arithx: That's fair, the base panic message will probably be sufficient.

PR #53, opened 2019-10-16
add filesystems.with_mount_unit

ajeddeloh:
Depends on https://github.com/coreos/vcontext/pull/10, will fix the go.mod change after that merges

Add a field to generate mount units automatically from the filesystems section. ~~Untested with FCOS.~~ Tested now, works.

Ended up doing some refactoring along the way.

Tests are failing because of the dependency on the vcontext PR


base/v0_2_exp/translate.go
```diff
@@ -100,3 +106,84 @@ func translateLink(from Link) (to types.Link, tm translate.TranslationSet) {
...
+			unitPath := path.New("json", "systemd", "units", len(ret.Systemd.Units))
+			ret.Systemd.Units = append(ret.Systemd.Units, newUnit)
+			ts.AddFromCommonSource(fromPath, unitPath, newUnit)
+		}
+	}
+	return ts
+}
+
+var (
+	unitTemplate = template.Must(template.New("unit").Parse(`# Generated by FCCT
```
arithx: Minors:
 - I generally prefer to have all module level units at the top of the individual source files right after the imports
 - Maybe name it `mountUnitTemplate`?

translate/set.go
```diff
@@ -0,0 +1,101 @@
...
+		str += fmt.Sprintf("%s: %v -> %v\n", k, v.From.String(), v.To.String())
+	}
+	return str
+}
+
+// AddTranslation adds a translation to the set
+func (ts TranslationSet) AddTranslation(from, to path.ContextPath) {
+	// create copies of the paths so if someone else changes from.Path the added translation does not change.
+	from.Path = append([]interface{}{}, from.Path...)
+	to.Path = append([]interface{}{}, to.Path...)
```
arithx: Couldn't these make use of the `ContextPath.Copy()` function?

ajeddeloh:
Tested with FCOS now, works.

base/v0_2_exp/schema.go
```diff
@@ -70,13 +70,14 @@ type FileContents struct {
...
-	Format         *string            `yaml:"format"`
-	Label          *string            `yaml:"label"`
-	Options        []FilesystemOption `yaml:"options"`
-	Path           *string            `yaml:"path"`
-	UUID           *string            `yaml:"uuid"`
-	WipeFilesystem *bool              `yaml:"wipe_filesystem"`
+	Device         string   `yaml:"device"`
+	Format         *string  `yaml:"format"`
+	Label          *string  `yaml:"label"`
+	Options        []string `yaml:"options"`
```
arithx: Should probably note at least in the commit message why this is changing to a `[]string` from `[]FilesystemOption`, also after a quick grep that would leave the `FilesystemOption` type alias unused so it can probably be dropped.

translate/translate.go
```diff
@@ -211,8 +211,27 @@ type TranslationSet struct {
...
+		FromTag: fromTag,
+		ToTag:   toTag,
+		Set:     map[string]Translation{},
+	}
+}
+
+func (ts TranslationSet) String() string {
+	str := fmt.Sprintf("from: %v\nto: %v\n", ts.FromTag, ts.ToTag)
+	for k, v := range ts.Set {
+		str += fmt.Sprintf("%s: %v -> %v\n", k, v.From.String(), v.To.String())
```
arithx: Optional: Might be useful to tab these entries so you know they belong to this specific `TranslationSet`.

ajeddeloh: I think I'll pass. That implies their under the from/to tags which isn't really accurate. It's only added for debugging so I don't think it's too important.

ajeddeloh:
Updated commit message, broke out go.* changes into their own commit.

ajeddeloh:
1) Will add
2) Yeah, those kept happening as I went because I got annoyed. I imagine that problem will also get better as the codebase matures
3) I don't. This is a shortcut for the common case. If users need something more they can either write their own unit or use drop ins. That said, once https://github.com/coreos/ignition/pull/874 merges we could propagate options from there as well, though that will only help for actual options, not the systemd extensions.

docs/configuration-v1_1-exp.md
```diff
@@ -56,6 +56,7 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_label_** (string): the label of the filesystem.
     * **_uuid_** (string): the uuid of the filesystem.
     * **_options_** (list of strings): any additional options to be passed to the format-specific mkfs utility.
+    * **_with_mount_unit_** (bool): Whether to generate a generic mount unit for this filesystem as well. If a more specific unit is needed, a custom one can be specified in the `systemd.units` section. The unit will be named with the [escaped][systemd-escape] version of the `path`.
```
lucab: s/bool/boolean/ (for consistency with other entries).
Also, this would benefit from an additional "Defaults to false" sentence.

lucab:
I'm sorry to be dragging this further, but I'd like to keep the focus a bit more on the point 3 above. I'd prefer if we could future-proof the schema a bit more to easily accommodate those (if we feel like, in the future).

What about tweaking your current schema a bit so that longer term it may look something like this:
 * `mount` (object, optional)
   * `create_unit` (boolean, optional, default: false)
   * `initrd` (boolean, optional, default: false)
   * `options` (list of string, optional, default: nil) <- this will be forwarded to both ignition and mount-unit

ajeddeloh:
For the examples you specified, options would be covered by https://github.com/coreos/ignition/pull/874 I'd think. Given that our initrd is static, what would be the reason for mounting it in the initrd?

ajeddeloh:
Also, note we can change the format later if need be by bumping major versions.

lucab:
> For the examples you specified, options would be covered by coreos/ignition#874 I'd think.

Yes. This is just to make sure we can forward from `fcct` to both Ignition and systemd.

> Given that our initrd is static, what would be the reason for mounting it in the initrd?

See https://github.com/coreos/rpm-ostree/issues/1936#issuecomment-549298710. The main usecase I'm thinking of is `/var`.

ajeddeloh:
What do you think of merging as is, but changing it in the future. It's still in the experimental spec, so we could do that without even changing version. I  _think_ the initrd case is probably esoteric enough we can also ask users to write their own configuration in that case.

lucab:
Ack, I trust your call on that (I don't have a good measure of how many common usecases would require that). Code LGTM.

Issue #69, opened 2020-01-25
FCCT doesn't complain about extra command-line arguments


gabx:
My **.yaml** configuration file is named `tth.fcc` and has been written using visual studio and the red hat yaml plugin. When saving the file, visual studio does not split any error. Everything is made from my home directory. I use fcct v 0.3.

```
% ls -al 
....
tth.yaml
% mv tth.yaml tth.fcc
% ls -al 
...
tth.fcc
%podman pull quay.io/coreos/fcct:v0.3.0
%podman run -i --rm quay.io/coreos/fcct:v0.3.0 -pretty -strict tth.fcc transpiled_config.ign
_
```

Nothing happens. The command outputs nothing and after a few minutes, I have to **ctr+c** to quit.
Now:

```
%podman run --rm -v /var/home/gabx/tth.fcc:/config.fcc:z quay.io/coreos/fcct:v0.3.0 -pretty -strict -input /config.fcc > transpiled_config.ign
%ls -al
....
transpiled_config.ign
tth.fcc
```

The command succeeded and I have a proper ignition file. 
No idea why one succeed and not the other one.

jlebon:
```
%podman run -i --rm quay.io/coreos/fcct:v0.3.0 -pretty -strict tth.fcc transpiled_config.ign
```

This isn't using stdin/stdout though, right? See the example invocation in https://github.com/coreos/fcct/blob/master/docs/getting-started.md#container:

```
podman run -i --rm quay.io/coreos/fcct:release -pretty -strict < your_config.fcc > transpiled_config.ign
```

gabx:
so you are telling me the correct command shall place the **.fcc** config in between less/greater symbols **<  >** ? 
I was wondering why these two symbols were placed here. I have been confused.

jlebon:
Yup, exactly! The `<` tells bash to feed the contents of the file specified to the command and `>` tells bash to write the output of the command to the file specified. Check out this page from the Bash manual for more info: https://www.gnu.org/software/bash/manual/html_node/Redirections.html.

bgilbert:
Retitled to reflect what I think is the actual bug here.  It's unusual that FCCT requires an `-input` option for specifying the input file, and if there's a bare filename passed on the command line, FCCT silently ignores it and waits for stdin.

We should fix the latter part, at least.  What would folks think about also deprecating `-input` and reading a bare filename arg instead?

jlebon:
[...]

PR #70, opened 2020-01-29
main: Switch from flag to pflag

bgilbert:
`pflag` is more flexible.  It allows us to add short options, which we do here.  It also supports hidden options, which is useful for fixing #69.

This is a backward-incompatible change.  `flag` accepts e.g. both `-strict` and `--strict`, but documents `-strict` in the help text.  `pflag` documents `--strict` and accepts only `--strict`.

This will break people's scripts.  However, we're technically still in 0.x-land.  If we want to switch to a more conventional parser with double-dash long arguments, it's only going to get harder from here.

Thoughts?


lucab:
Breaking is not nice, but I'm in favor of doing this cleanup. Thus, better now than later.

I'm not exactly buying the 0.x argument, but I think at this point most of `fcct` users are still in the initial manual/exploration phase OR pinning the container version. So the blast radius of this PR should be not too big.

ashcrow:
> Breaking is not nice, but I'm in favor of doing this cleanup. Thus, better now than later.

I agree :+1:

ashcrow:
4 :+1:'s ... merging :smile:

lucab:
@jlebon that's https://github.com/coreos/fedora-coreos-docs/pull/35.

dustymabe:
since this merged should we consider a larger version bump for our next release?

bgilbert:
@dustymabe SemVer says anything goes for 0.x versions.  We'll call out the breaking change in the release notes.

PR #72, opened 2020-01-29
main: accept input filename as bare argument; deprecate --input

### LINKED ISSUE #69: FCCT doesn't complain about extra command-line arguments

bgilbert:
Accept input filename as bare argument rather than requiring the `--input` option.  Continue accepting `--input` but hide it from `--help`.

Also, fail if we receive extra command-line arguments.

Fixes #69.  Depends on #70.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -44,9 +44,24 @@ func main() {
 	pflag.BoolVarP(&options.Strict, "strict", "s", false, "fail on any warning")
 	pflag.BoolVarP(&options.Pretty, "pretty", "p", false, "output formatted json")
 	pflag.StringVar(&input, "input", "", "read from input file instead of stdin")
+	pflag.Lookup("input").Deprecated = "specify filename directly on command line"
+	pflag.Lookup("input").Hidden = true
 	pflag.StringVarP(&output, "output", "o", "", "write to output file instead of stdout")
 
+	pflag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: %s [options] [input-file]\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "Options:\n")
+		pflag.PrintDefaults()
+	}
 	pflag.Parse()
+
+	args := pflag.Args()
+	if len(args) == 1 && input == "" {
+		input = args[0]
+	} else if len(args) > 0 {
+		pflag.Usage()
+		os.Exit(2)
+	}
 
 	if helpFlag {
 		pflag.Usage()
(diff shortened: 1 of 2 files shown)


ashcrow:
https://github.com/coreos/fcct/pull/70 has merged.

bgilbert:
Rebased.

internal/main.go
```diff
@@ -43,11 +43,25 @@ func main() {
 	pflag.BoolVarP(&versionFlag, "version", "V", false, "print the version and exit")
 	pflag.BoolVarP(&options.Strict, "strict", "s", false, "fail on any warning")
 	pflag.BoolVarP(&options.Pretty, "pretty", "p", false, "output formatted json")
-	pflag.StringVar(&input, "input", "", "read from input file instead of stdin")
+	pflag.StringVar(&input, "input", "", "deprecated: read from input file instead of stdin")
+	pflag.Lookup("input").Hidden = true
```
jlebon: Note there's a `MarkDeprecated` we could use instead that hides it and also emits a deprecated message when used: https://github.com/coreos/mantle/pull/1151/commits/d9fb0d258cffcfe136ccc0e15e4c6476c500e754

bgilbert: `Deprecated` doesn't itself hide the option; we need both that and `Hidden`.

bgilbert:
Updated!

bgilbert:
Ah, sorry, I wasn't clear.  The last update sets both `Deprecated` and `Hidden` and that seems to work fine.

internal/main.go
jlebon: Hmm, weird. That's not what I'm seeing here:

```
{pet ~}$ plume pre-release --help |& grep channel
{pet ~}$ plume pre-release --help |& grep stream
  -S, --stream string             target stream (default "testing")
{pet ~}$ plume pre-release --channel foo --help |& grep stream
Flag --channel has been deprecated, use --stream instead
  -S, --stream string             target stream (default "testing")
```

bgilbert: Not sure.  `cobra` may drive `pflags` in a different way.  Anyway, the current arrangement seems to work.

Issue #78, opened 2020-02-13
minimize the generated Ignition json


dustymabe:
It would be nice if we don't put extraneous sections in our Ignition config that might cause confusion.

For example:

```
[dustymabe@media tmp]$ cat foo.yaml
variant: fcos
version: 1.0.0
passwd:
  users:
    - name: dusty
      ssh_authorized_keys:
        - ssh-rsa AAAAxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx foo@bar.com
[dustymabe@media tmp]$ 
[dustymabe@media tmp]$ cat foo.yaml | fcct --pretty
{
  "ignition": {
    "config": {
      "replace": {
        "source": null,
        "verification": {}
      }
    },
    "security": {
      "tls": {}
    },
    "timeouts": {},
    "version": "3.0.0"
  },
  "passwd": {
    "users": [
      {
        "name": "dusty",
        "sshAuthorizedKeys": [
[...]

PR #90, opened 2020-04-03
config/common: Exclude empty structs from generated Ignition json

LorbusChris:
The  "github.com/clarketm/json"  package supports zero values of structs
with omitempty markers when marshaling them.
See [1] for context on this.

Closes: #78

[1]  golang/go#11939


bgilbert:
This does indeed remove the extraneous attributes.

@jlebon @lucab How do you feel about this approach?

lucab:
Not nice to crawl away from Go stdlib, but not the first time we have to do that anyway. The upstream here looks serious enough.

I don't have any better ideas. An alternative could be turning all non-required structs into references in Ignition. I guess that would be a major semver break, which sounds worse that this.

Unless we find other better approaches, I'd go with this.

bgilbert:
> I don't have any better ideas. An alternative could be turning all non-required structs into references in Ignition.

I believe Andrew made a conscious decision for spec 3 to standardize on not using references to structs.  I don't recall the details, though.

> Unless we find other better approaches, I'd go with this.

In principle we could post-process the output by re-parsing it into `map[string]interface{}`, walk it alongside the struct definitions, and delete any field that matches the zero value and has `omitempty` in the struct tag.  I don't know that that's better, though.

LorbusChris:
I assume this is the option with least local code changes at least ;)
Other projects use this library as well, e.g. linkerd2: https://github.com/linkerd/linkerd2/commit/b94e03a8a6c0d71527e2cb9893ba63b30ae0fbd0

jlebon:
Yeah... this is messy without a proper `Option<T>` like in Rust. The closest thing I think is indeed using pointers to structs.

No better ideas either. Using that package seems reasonable if it solves the issue. At least we know we can ship with that for now and fall back to reflection in the future if we have to drop it for some reason.

> I believe Andrew made a conscious decision for spec 3 to standardize on not using references to structs. I don't recall the details, though.

I'd be interested in more info on this if anyone else recalls. Cosa for example does this now: https://github.com/coreos/coreos-assembler/pull/1239 (though it's of course not a public API in the same way).

Ignition does at least have field that are pointers to basic datatypes. E.g. `httpResponseHeaders` is of type `*int`: https://github.com/coreos/ignition/blob/f73e34b2c21c1415eef9344d32f36ebfe90e30eb/config/v3_1_experimental/types/schema.go#L209 because "not provided" is not the same as 0.

Without using pointers for structs too, ISTM we're artificially constraining our schemas to have "missing struct" being equivalent to "present and equal to zero value", which might not fit well at all with the functionality in question.

Anyway agreed this would require a major semver bump anyway, which doesn't seem worth it.

bgilbert:
I think the idea was roughly that struct embedding resulted in simpler code.  Outside of an array, there's no case in the spec where the presence/absence of a zero-valued struct makes a semantic difference.

PR #91, opened 2020-04-04
base/v0_2_exp: translate httpHeaders fields

bgilbert:


lucab:
[Please bear with me if the following is a dumb question]
Don't we have to wait for `3.1.0` Ignition schema in order to wire this here? Otherwise, won't this generate a `-experimental` version which eventually won't be accepted anymore by Ignition?

bgilbert:
FCCs are versioned.  Base `0.2-exp` is only used by FCCL `1.1.0-experimental`, which targets Ignition spec `3.1.0-experimental`.  FCCL `1.0.0` targets spec `3.0.0` and isn't affected by this PR.

lucab:
Yes, I'm only looking at the `1.1.0-experimental` -> `3.1.0-experimental` edge, which produces an output which is valid today but it won't be at an arbitrary point in the future.

But I take that this is already the case today, so overall it is not deemed a concern, right?
Other than this silly doubt of mine (about an already existing behavior), the PR looks fine.

bgilbert:
> Yes, I'm only looking at the `1.1.0-experimental` -> `3.1.0-experimental` edge, which produces an output which is valid today but it won't be at an arbitrary point in the future.

Yup, that's already the case, and is by design.

lucab:
LGTM, thanks for the concise/explicit walkthrough!

PR #93, opened 2020-04-06
config/common: validate duplicate keys

lucab:
This unifies the final validation step on the generated JSON config,
adding a duplicate check on keys which was previously missing and
only performed by Ignition.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/451

### DIFF
--- a/config/v1_0/fcos.go
+++ b/config/v1_0/fcos.go
@@ -77,10 +77,17 @@ func TranslateBytes(input []byte, options common.TranslateOptions) ([]byte, repo
 		return nil, r, err
 	}
 
-	second := validate.Validate(final, "json")
-	common.TranslateReportPaths(&second, translations)
-	second.Correlate(contextTree)
-	r.Merge(second)
+	// Check for invalid duplicated keys.
+	dupsReport := validate.ValidateCustom(final, "json", ignvalidate.ValidateDups)
+	common.TranslateReportPaths(&dupsReport, translations)
+	dupsReport.Correlate(contextTree)
+	r.Merge(dupsReport)
+
+	// Validate JSON semantics.
+	jsonReport := validate.Validate(final, "json")
+	common.TranslateReportPaths(&jsonReport, translations)
+	jsonReport.Correlate(contextTree)
+	r.Merge(jsonReport)
 
 	if r.IsFatal() {
 		return nil, r, common.ErrInvalidGeneratedConfig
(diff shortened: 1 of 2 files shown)


config/common/common.go
```diff
@@ -95,3 +97,18 @@ func TranslateReportPaths(r *report.Report, ts translate.TranslationSet) {
 		r.Entries[i].Context = context
 	}
 }
+
+// ValidateTranslated validates the final translated JSON config.
+func ValidateTranslated(r *report.Report, translated interface{}, contextTree tree.Node, ts translate.TranslationSet) {
```
bgilbert, line 102: The general pattern seems to be to return a `Report` and have the caller merge it.

We're open-coding the rest of `TranslateBytes` within each config version, so it seems a bit odd to abstract out this specific part.

config/common/common.go
```diff
@@ -95,3 +97,18 @@ func TranslateReportPaths(r *report.Report, ts translate.TranslationSet) {
...
+	// Check for invalid duplicated keys.
+	dupsReport := validate.ValidateCustom(translated, "json", ignvalidate.ValidateDups)
```
bgilbert, line 104: Self note: Ignition's `ValidateWithContext` utility function (which `ValidateTranslated` is partially reimplementing) calls `ValidateDups` _after_ `validate.Validate` but it probably doesn't matter.

config/common/common.go
lucab: I pushed a rebased version, but I'm not sure if your last sentence meant you'd prefer to get rid of this function and just have these few lines duplicated in the two callers.

bgilbert: Without a larger refactoring, I think I'd be in favor of duplicating the lines in the callers.  But I won't block on it.

config/common/common.go
```diff
@@ -95,3 +97,22 @@ func TranslateReportPaths(r *report.Report, ts translate.TranslationSet) {
 		r.Entries[i].Context = context
 	}
 }
+
+// ValidateTranslated validates the final translated JSON config.
+func ValidateTranslated(translated interface{}, contextTree tree.Node, ts translate.TranslationSet) report.Report {
+	outReport := report.Report{}
```
bgilbert, line 103: nit: Convention is to start from the first report returned by a callee.

config/common/common.go
lucab: There is possibly a larger rework to be done on Ignition side (I'm not sure why `ValidateWithContext` sidesteps the normal validation entrypoint), but I honestly got a bit lost in validation indirection layers trying to figure out where to best plug it.

lucab:
I've split back and duplicated the logic in the two callers. PTAL.

dustymabe:
trivial LGTM

PR #97, opened 2020-04-30
Update 1.1.0-exp for Resource type and SHA-256 verification support

bgilbert:


docs/configuration-v1_1-exp.md
```diff
@@ -14,14 +14,14 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
         * **name** (string): the header name.
         * **_value_** (string): the header contents.
       * **_verification_** (object): options related to the verification of the config.
-        * **_hash_** (string): the hash of the config, in the form `<type>-<value>` where type is `sha512`.
+        * **_hash_** (string): the hash of the config, in the form `<type>-<value>` where type is either `sha512` or `sha256`.
```
arithx, line 17: Is there anything preventing the use of `inline` for the old `ConfigReference` or `CaReference` portions or should that also be documented now?

bgilbert: Whoops, good catch.  Updated.

PR #98, opened 2020-05-01
base/v0_2: include mount options in generated mount unit

bgilbert:
Any mount options that should be applied by Ignition are probably relevant in the real root too.  If we're asked to generate a mount unit, copy them over.


ashcrow:
LGTM. Merging.

PR #99, opened 2020-05-05
base/v0_2: add sugar to embed Resource from a local file

bgilbert:
If `local` is specified instead of `source` or `inline`, read a local file from a base directory and encode it into a data URI.  We require the base directory to be specified on the command line, so that transpiling untrusted configs doesn't leak information from the host system by default.

This required plumbing an options struct into custom translators, and plumbing error returns (via a `report.Report`) back out.

Fixes https://github.com/coreos/fcct/issues/83.


config/common/common.go
```diff
@@ -33,9 +35,14 @@ var (
...
 )
 
+type BaseOptions = base.TranslateOptions
+type FCOSOptions = fcos.TranslateOptions
+
 type TranslateOptions struct {
 	Pretty bool
 	Strict bool
+	BaseOptions
+	FCOSOptions
```
lucab, line 45: I think this uninhabited and currently unused, embedded here only for future usage. Thus I'm wondering, how do you plan to wire distro-specific options to the CLI? Wouldn't it be enough to have a single set of options, mirroring what the CLI exposes (and only using the relevant subset in each distro)?

I know this fully-customizable approach looks enticing, but I'm wary of inducing a mess at the CLI-flags level.

bgilbert: I was expecting all options to be wired into the CLI equally, without distinguishing between distro-specific and generic ones.  You're right that that might have been confusing.  I've dropped the distro-specific options struct.

base/v0_2_exp/translate.go
```diff
@@ -100,6 +107,53 @@ func translateResource(from Resource, options base.TranslateOptions) (to types.R
...
+			return
+		}
+
+		infile, err := os.Open(filePath)
+		if err != nil {
+			r.AddOnError(c, err)
+			return
+		}
+		defer infile.Close()
+		contents, err := ioutil.ReadAll(infile)
```
arithx, line 143: Is there a particular reason to be opening the file manually and performing a `ReadAll` rather than a `ReadFile` call?

docs/configuration-v1_1-exp.md
```diff
@@ -9,16 +9,18 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
 * **ignition** (object): metadata about the configuration itself.
   * **_config_** (objects): options related to the configuration.
     * **_merge_** (list of objects): a list of the configs to be merged to the current config.
-      * **_source_** (string): the URL of the config. Supported schemes are `http`, `https`, `s3`, `tftp`, and [`data`][rfc2397]. Note: When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified. Mutually exclusive with `inline`.
-      * **_inline_** (string): the contents of the config. Mutually exclusive with `source`.
+      * **_source_** (string): the URL of the config. Supported schemes are `http`, `https`, `s3`, `tftp`, and [`data`][rfc2397]. Note: When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified. Mutually exclusive with `inline` and `local`.
+      * **_inline_** (string): the contents of the config. Mutually exclusive with `source` and `local`.
+      * **_local_** (string): a local path to the contents of the config. Mutually exclusive with `source` and `inline`.
```
arithx, line 14: Might be worth adding a mention to the `-d/--files-dir` requirement in the specification.

base/v0_2_exp/translate.go
bgilbert: Nope.  I haven't written Go in a while.  :slightly_smiling_face:

bgilbert:
Updated.

joyartoun:
Hi!
I am dependent on a release of fcct to be able to use this feature. Do you know when there will be a release?

/J

bgilbert:
@joyartoun Hopefully within the next couple weeks.

PR #100, opened 2020-05-05
base/v0_2: auto-select encoding for inline/local file contents

bgilbert:
Use base64 encoding if the result is smaller than URL escaping.  If the user didn't already enable compression, and base64-encoded gzip is smaller than the alternatives, use that instead.

All versions of Ignition that support spec 3.1.0 should support compression in data URLs.  Older versions don't, so limit this to base 0.2.

Depends on https://github.com/coreos/fcct/pull/99.  Fixes https://github.com/coreos/fcct/issues/94.

### DIFF
--- a/base/v0_2_exp/translate_test.go
+++ b/base/v0_2_exp/translate_test.go
@@ -64,23 +64,28 @@ func verifyTranslations(set translate.TranslationSet, exceptions ...translate.Tr
 
 // TestTranslateFile tests translating the ct storage.files.[i] entries to ignition storage.files.[i] entries.
 func TestTranslateFile(t *testing.T) {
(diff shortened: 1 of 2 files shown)


bgilbert:
Actually, the better approach is to automatically choose the smallest of three encodings: plain, base64, or gzip+base64.  Will update.

base/v0_2_exp/translate_test.go
```diff
@@ -339,7 +348,88 @@ func TestTranslateFile(t *testing.T) {
...
+				},
+				Append: []Resource{
+					{
+						// should compress
+						Local: util.StrToPtr("file-2"),
+					},
+					{
+						// should not compress
+						Local:       util.StrToPtr("file-2"),
+						Compression: util.StrToPtr("gzip"),
```
arithx, line 372: If I'm reading this right both this case & the one below shouldn't be compressed because they are specifying the compression. If so it might be more obvious to choose a different compression string to make it more obvious at a glance.

bgilbert:
Updated.

PR #101, opened 2020-05-07
Switch to final version of Ignition spec 3.1.0

bgilbert:
Switch FCC spec 1.1.0-experimental to produce stable spec 3.1.0 configs.  This breaks 1.1.0-experimental compatibility with Ignition < 2.3.0, which is permissible because we can break experimental FCC specs at any time.  In practice, https://github.com/coreos/fcct/pull/100 breaks `inline` in 1.1.0-experimental anyway for older versions of Ignition.

We'll support 3.2.0-experimental in 1.2.0-experimental, once 1.1.0 is stabilized.


jlebon:
Can we write down the semantics of how FCC spec versioning vs Ignition spec versioning match up in a document under `docs/`? (Or if it's already written somewhere else, let's put it under there?).

bgilbert:
@jlebon There's no general rule.  In each FCC spec doc, the description of the `version` attribute gives the corresponding Ignition spec version ([example](https://github.com/coreos/fcct/blob/master/docs/configuration-v1_1-exp.md)).

PR #102, opened 2020-05-09
base/v0_2: support inclusion of local directory trees

bgilbert:
Fixes https://github.com/coreos/fcct/issues/76.

I'm interested in feedback on whether `follow_symlinks` is a good idea.  It seems a bit odd to me to change  symlink behavior at the level of an entire tree, and I'm not sure whether there's enough of a use case to justify the feature.


ashcrow:
cc'ing @christianh814 too

base/v0_2_exp/translate.go
```diff
@@ -88,6 +95,23 @@ func translateIgnition(from Ignition, options base.TranslateOptions) (to types.I
 	return
 }
 
+func translateStorage(from Storage, options base.TranslateOptions) (to types.Storage, tm translate.TranslationSet, r report.Report) {
```
arithx, line 98: It doesn't feel great that we have to define a custom translator to just ignore a field that only exists in one struct. I get that we don't want to automatically silently ignore fields but maybe we can add another mechanism to explicitly say "we're handling this field elsewhere, please ignore it" when the general translator is passing through?

base/v0_2_exp/validate_test.go
```diff
@@ -134,3 +134,25 @@ func TestValidateResource(t *testing.T) {
...
+		},
+	}
+
+	for i, test := range tests {
+		actual := test.in.Validate(path.New("yaml"))
+		expected := report.Report{}
+		expected.AddOnError(path.New("yaml"), test.out)
+
+		if !reflect.DeepEqual(actual, expected) {
+			t.Errorf("#%d: expected %+v got %+v", i, expected, actual)
```
arithx, line 155: Should this not be using the `%v` formatting with calls to `format` on both `expected` & `actual` similar to what was done in one of the previous commits?

bgilbert:
Discussed feature scope OOB with @arithx, @cgwalters, @lucab, and @jlebon.  There's a rough consensus to provide basic functionality for now, with reasonable defaults, and defer additional functionality until later.  The existing `follow_symlinks` field is an unusual use case and adds complexity, so I'll remove it from this PR.  Taking file modes from the source tree is incompatible with keeping trees in Git, since Git only tracks the execute bit, so I'll change this PR to take only the execute bit from the tree.

Deferred functionality includes tree-level settings to:

- Override default user/group for the tree
  - Override functionality will require FCCT to create `directory` nodes (conditionally, to minimize Ignition config size)
  - Detection of per-node overrides requires some care; if an override specifies `user.id` we shouldn't apply a tree-level `user.name`
- Override default file/directory modes
  - E.g. `default_file_mode` and `default_directory_mode`, where the file mode is modified by the execute bit
- Set `overwrite` field
  - E.g. `overwrite_files_links`, skipping directories, since setting `overwrite` on directories will delete their contents
- Follow symlinks
  - This requires manually recursing into directory symlinks, since `filepath.Walk()` won't do it, and maintaining a set of visited nodes to detect infinite recursion

bgilbert:
Updated to drop `follow_symlinks`, set the output file mode to 0644 or 0755 depending on the execute bit of the source file, and address comments.

PR #107, opened 2020-05-27
Stabilize spec 1.1.0; create spec 1.2.0-experimental

bgilbert:


lucab:
LGTM, but leaving some space in case @arithx was planning to look at this.

I think it would be good (for both author and reviewers) to have a SOP document like we do for "stabilizing spec in Ignition" to follow along.

bgilbert:
Okay, will add the SOP in a followup.

PR #112, opened 2020-06-16
base/v0_3_exp: add LUKS

arithx:
Adds the LUKS related structs as well as updating the `with_mount_unit` logic
to target `remote-fs.target` when creating mount units for Tang based LUKS
devices


arithx:
Draft PR while waiting for Ignition to land so the vendor can be updated. Tested locally via doing a vendor replace (`go mod edit -replace github.com/coreos/ignition/v2=/path/to/local/ignition`)

base/v0_3_exp/translate.go
```diff
@@ -387,7 +409,20 @@ func (c Config) addMountUnits(ret *types.Config) translate.TranslationSet {
 			continue
 		}
 		fromPath := path.New("yaml", "storage", "filesystems", i, "with_mount_unit")
-		newUnit := mountUnitFromFS(fs)
+		remote := false
+		// check filesystems targeting /dev/mapper devices against LUKS to determine if a
+		// remote mount is needed
+		if strings.HasPrefix(fs.Device, "/dev/mapper/") {
```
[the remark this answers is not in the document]
arithx: Hmm, can use `/dev/disk/by-id/dm-name-<name>`

arithx:
Marked ready for review. Updated to bump Ignition to the 2.5.0 release & the state of LUKS in that release.

base/v0_3_exp/translate.go
```diff
@@ -387,7 +423,21 @@ func (c Config) addMountUnits(ret *types.Config) translate.TranslationSet {
 			continue
 		}
 		fromPath := path.New("yaml", "storage", "filesystems", i, "with_mount_unit")
-		newUnit := mountUnitFromFS(fs)
+		remote := false
+		// check filesystems targeting /dev/mapper devices against LUKS to determine if a
+		// remote mount is needed
+		if strings.HasPrefix(fs.Device, "/dev/mapper/") || strings.HasPrefix(fs.Device, "/dev/disk/by-id/dm-name-") {
```
jlebon, line 429: Hmm, wish there was a more elegant way to do this, but I don't have a better idea. Can we at least document this in `with_mount_unit`?

base/v0_3_exp/translate.go
```diff
@@ -387,7 +423,21 @@ func (c Config) addMountUnits(ret *types.Config) translate.TranslationSet {
...
+			for _, luks := range c.Storage.Luks {
+				// LUKS devices are opened with their name specified
+				deviceName := filepath.Base(fs.Device)
```
bgilbert, line 432: Can be hoisted outside the loop.

As @LorbusChris noticed, we should be using `path` instead of `filepath` when we're referring to specifically Unix paths.  Otherwise we're expecting `C:\dev\mapper\my-luks-device` on Windows.

base/v0_3_exp/translate.go
```diff
@@ -387,7 +423,21 @@ func (c Config) addMountUnits(ret *types.Config) translate.TranslationSet {
...
+				if deviceName == luks.Name || deviceName == fmt.Sprintf("dm-name-%s", luks.Name) {
```
bgilbert, line 433: This throws away information (whether to expect a `dm-name-` prefix) and then tries to add it back.  In the _extremely unlikely case_ that someone creates LUKS devices named `foo` and `dm-name-foo`, it will get confused.  Why not flip it around?

```suggestion
				if fs.Device == fmt.Sprintf("/dev/mapper/%s", luks.Name) || fs.Device == fmt.Sprintf("/dev/disk/by-id/dm-name-%s", luks.Name) {
```

base/v0_3_exp/translate.go
```diff
@@ -387,7 +423,21 @@ func (c Config) addMountUnits(ret *types.Config) translate.TranslationSet {
...
+					if luks.Clevis != nil && len(luks.Clevis.Tang) > 0 {
+						remote = true
```
bgilbert, line 435: `break`

docs/configuration-v1_2-exp.md
```diff
@@ -124,6 +124,28 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
       * **_name_** (string): the group name of the owner.
     * **target** (string): the target path of the link
     * **_hard_** (boolean): a symbolic link is created if this is false, a hard one if this is true.
+ * **_luks_** (list of objects): the list of luks devices to be created. Every device must have a unique `device` & `name`.
```
bgilbert, line 127: Please sync this with the current contents of the Ignition 3.2-exp spec.

docs/configuration-v1_2-exp.md
```diff
@@ -124,6 +124,28 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
...
+    * **name** (string): the name of the luks device. Every device must have a unique `name`.
+    * **_device_** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
+    * **_key_file_** (string): options related to the contents of the key file.
+      * **_compression_** (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
+      * **_source_** (string): the URL of the file contents. Supported schemes are `http`, `https`, `tftp`, `s3`, and [`data`][rfc2397]. When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified. If source is omitted and a regular file already exists at the path, Ignition will do nothing. If source is omitted and no file exists, an empty file will be created. Mutually exclusive with `inline` and `local`.
+      * **_inline_** (string): the contents of the file. Mutually exclusive with `source` and `local`.
```
bgilbert, line 133: s/file/key file/.  Also in `local`.

base/v0_3_exp/translate.go
```diff
@@ -237,6 +260,19 @@ func translateLink(from Link, options base.TranslateOptions) (to types.Link, tm
 	return
 }
 
+func translateLuks(from Luks, options base.TranslateOptions) (to types.Luks, tm translate.TranslationSet, r report.Report) {
```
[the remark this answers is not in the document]
arithx: Nope, `key_file` is a `Resource` object so without it the binary panics during translation.

base/v0_3_exp/translate.go
arithx: Ended up dropping as a result of going with the other suggestion for checking the entire device path rather than just the device name.

arithx:
Updated

base/v0_3_exp/translate.go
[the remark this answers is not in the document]
arithx: Actually adding the custom translator for resources to the `ToIgn3_2` function suffices, dropped the function.

PR #124, opened 2020-08-05
base: Make gzip auto-compression skipable

LorbusChris:
Fixes: https://github.com/coreos/fcct/issues/123

If this is agreeable, can we also port it back to v0.2?

### DIFF
--- a/base/options.go
+++ b/base/options.go
@@ -15,5 +15,6 @@
 package base
 
 type TranslateOptions struct {
-	FilesDir string // allow embedding local files relative to this directory
+	FilesDir                  string // allow embedding local files relative to this directory
+	NoResourceAutoCompression bool   // skip automatic compression of inline/local resources
 }
--- a/base/v0_2/translate.go
+++ b/base/v0_2/translate.go
@@ -140,7 +140,7 @@ func translateResource(from Resource, options base.TranslateOptions) (to types.R
 			return
 		}
 
-		src, gzipped, err := makeDataURL(contents, to.Compression)
+		src, gzipped, err := makeDataURL(contents, to.Compression, options.NoResourceAutoCompression)
 		if err != nil {
 			r.AddOnError(c, err)
 			return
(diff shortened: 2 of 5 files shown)


base/v0_3_exp/translate_test.go
```diff
@@ -255,6 +255,7 @@ func TestTranslateFile(t *testing.T) {
 			File{
 				Path: "/foo",
 				Contents: Resource{
+					// String is too short for auto gzip compresison
```
bgilbert, line 258:
```suggestion
					// String is too short for auto gzip compression
```

base/options.go
```diff
@@ -15,5 +15,6 @@
 package base
 
 type TranslateOptions struct {
-	FilesDir string // allow embedding local files relative to this directory
+	FilesDir                   string // allow embedding local files relative to this directory
+	DisableGzipAutoCompression bool   // config source compression on Ignition output
```
bgilbert, line 19:
```suggestion
	NoResourceAutoCompression bool   // skip automatic compression of inline/local resources
```

bgilbert:
Yeah, we can add this to 0.2 also.  It's not a user-visible spec change.

LorbusChris:
@bgilbert thanks! I've fixed the typo, renamed the bool and also added it to 0.2.

base/v0_2/translate.go
```diff
@@ -171,7 +171,7 @@ func translateResource(from Resource, options base.TranslateOptions) (to types.R
 	return
 }
 
-func makeDataURL(contents []byte, currentCompression *string) (uri string, gzipped bool, err error) {
+func makeDataURL(contents []byte, currentCompression *string, NoResourceAutoCompression bool) (uri string, gzipped bool, err error) {
```
arithx, line 174: minor: variables should be camelCase

LorbusChris: fixed

PR #140, opened 2020-10-20
Password Hashing Documentation Updates

tcarrio:
- Includes shared link for password-hashes
- Updates references for `password_hash` to go to RH Knowledge Base
- RH Knowledge Base article covers all supported password schemes

I found myself having to dig through documentation around this since it wasn't referenced here. I figure that the Red Hat Knowledge Base is a fair reference, and it covers all supported hash functions, the format of the hash string, and more. This was useful for me while I was setting up some FCC files, figured I'd contribute it back up for anyone else who might need it.

Let me know if there's anything else I should include.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -45,6 +45,40 @@ passwd:
       shell: /bin/bash
 ```
 
+### Using Password Authentication
+
+You can use a Fedora CoreOS Config to set a password for a local user. Building on the previous example, we can configure the `password_hash` for one or more users:
+
+<!-- fedora-coreos-config -->
+```yaml
+variant: fcos
+version: 1.1.0
+passwd:
+  users:
+    - name: user1
+      ssh_authorized_keys:
+        - key1
+      password_hash: $y$j9T$aUmgEDoFIDPhGxEe2FUjc/$C5A...
+      home_dir: /home/user1
+      no_create_home: true
+      groups:
+        - wheel
+        - plugdev
+      shell: /bin/bash
+```
+
+To generate a secure password hash, use the `mkpasswd` command:
+
+```
+$ mkpasswd --method=yescrypt
+Password:
+$y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
+```
+
+The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
+
+For more information, see the Fedora CoreOS documentation on [Authentication][fcos-auth-docs].
+
 ## Storage and files
 
 ### Files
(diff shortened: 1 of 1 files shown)


bgilbert (dismissed):
Thanks for the PR!  FCOS has some documentation about password hashing [here](https://docs.fedoraproject.org/en-US/fedora-coreos/authentication/#_using_password_authentication).  We should probably add some documentation to FCCT as well, perhaps in the examples doc.  A couple notes:

1. We should either carry the docs ourselves or link to a more authoritative source.  That link shouldn't be in the config spec docs unless the reference is normative rather than informative.
2. That access.redhat.com article is paywalled, so isn't an appropriate reference here.

tcarrio:
@bgilbert That is a much better reference in my opinion. That is a great find. I'll update the links to point there instead. Thank you! 👍

bgilbert:
Per point 1 in https://github.com/coreos/fcct/pull/140#pullrequestreview-512312679, we shouldn't put that link directly in the config spec docs.  It might make sense to copy the linked section from the FCOS docs into a new section in `examples.md`.

tcarrio:
Oh I see, yeah that would work as well! I'll get about to that change tomorrow 👍

tcarrio:
Updated with an example showcasing how to use the `password_hash` for password authentication, as well as included a reference to the Fedora CoreOS documentation for Authentication. Let me know if you would like me to touch this up at all.

docs/examples.md
```diff
@@ -45,6 +45,39 @@ passwd:
       shell: /bin/bash
 ```
 
+### Using Password Authentication
+
+Fedora CoreOS ships with no default passwords. You can use a Fedora CoreOS Config to set a password for a local user. Building on the previous example, we can confgure the `password_hash` for one or more users:
```
bgilbert, line 50: I'm inclined to say that we shouldn't be documenting distro defaults in the tool docs, so I'd probably drop the first sentence.

spelling: `configure`

docs/examples.md
```diff
@@ -45,6 +45,39 @@ passwd:
...
+
+```yaml
```
bgilbert, line 52: Needs `<!-- fedora-coreos-config -->` header so CI will test the config.

docs/examples.md
bgilbert: I see your :+1: but no code change for this comment.  Did that part get missed?

tcarrio: I saw the review after the push, I'll be updating it shortly. Just wanted to indicate I saw it

bgilbert (approved):
Thanks for the PR!

PR #141, opened 2020-10-22
*: stabilize base/v0_3

arithx:
1. Bump Ignition to 2.7.0
1. Stabilize `base/v0_3`
1. Add `base/v0_4_exp`
1. Update docs


docs/migrating-configs.md
```diff
@@ -12,6 +12,104 @@ Occasionally, there are changes made to Fedora CoreOS configuration that break b
...
+    - path: /var/lib/tang_example
+      device: /dev/disk/by-id/dm-name-tang-example
+      format: ext4
+      label: TANG-EXAMPLE
+      with_mount_unit: true
+```
+
+### User/group deletion
+
+The `passwd` `users` and `groups` sections have a new field `shouldExist`. If specified and false, Ignition will delete the specified user or group if it exists.
```
bgilbert, line 82: `should_exist`

docs/migrating-configs.md
```diff
@@ -12,6 +12,104 @@ Occasionally, there are changes made to Fedora CoreOS configuration that break b
...
+
+<!-- fedora-coreos-config -->
+```yaml
+variant: fcos
+version: 1.2.0
+passwd:
+  users:
+    - name: core
+      should_exist: true
```
bgilbert, line 91: The corresponding Ignition example has these as `false`.  Why `true` here?  It's a no-op.

docs/migrating-configs.md
```diff
@@ -12,6 +12,104 @@ Occasionally, there are changes made to Fedora CoreOS configuration that break b
...
+  groups:
+    - name: core
+      should_exist: true
+```
+
+### GCS URL support
```
bgilbert, line 97: In hindsight, we should have spelled out "Google Cloud Storage" at least once in the source document.

base/v0_3/schema.go
```diff
@@ -27,6 +28,12 @@ type Config struct {
 	Systemd  Systemd  `yaml:"systemd"`
 }
 
+type Custom struct {
```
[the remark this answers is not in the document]
arithx: We could but we'd have to define a custom translator (Ignition also named the struct `Custom` and in [`translatableStruct`](https://github.com/coreos/fcct/blob/master/translate/translate.go#L72-L74) we require the name be the same for automatic translation).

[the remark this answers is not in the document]
bgilbert: I'm not sure it's worth a custom translator.  I'm okay leaving for now.

docs/migrating-configs.md
arithx: Went ahead and changed both occurrences.

arithx:
Updated

bgilbert (approved):

PR #156, opened 2020-11-14
base: warn if file/dir modes appear to have been specified in decimal

### LINKED ISSUE #131: Consider supporting 0644 or 644 interchangeably like chmod
When using `chmod` you can use 0644 or 644 interchangeably. The result is the same: 

```
$ ls -lah test.dhall; chmod 0644 test.dhall; ls -lah test.dhall
-rw-r--r--. 1 jdoss jdoss 193 Aug 29 23:28 test.dhall
-rw-r--r--. 1 jdoss jdoss 193 Aug 29 23:28 test.dhall
$ ls -lah test.dhall; chmod 644 test.dhall; ls -lah test.dhall
-rw-r--r--. 1 jdoss jdoss 193 Aug 29 23:28 test.dhall
-rw-r--r--. 1 jdoss jdoss 193 Aug 29 23:28 test.dhall
```
We should consider letting users write modes like this as well. 

```
variant: fcos
version: 1.1.0
storage:
  files:
    - path: /opt/file
      contents:
        inline: Hello, world!
      mode: 644
      user:
        id: 500
      group:
        id: 501
```

If the special bits are not set, just assume it's zero. I ran into this issue when trying to use [Dhall Lang](https://github.com/dhall-lang/dhall-lang) to manage my FCCs and Dhall will strip off leading zeros when you convert YAML to Dhall and then you end up with wonky permissions when you feed your Dhall back into YAML for consumption via FCCT. 

FWIW Dhall is looking into how to handle https://github.com/dhall-lang/dhall-lang/issues/1058#issuecomment-683375211 octal numerals on their end, but I feel that giving the same interchangeable user experience as `chmod` in a FCC would be a nice thing to have.

bgilbert:
Users sometimes forget to write file/directory modes with a leading zero and end up with a nonsensical configuration.  If a specified mode is unreasonable but would have been reasonable if it had been written in octal, report a warning.  Apply to all spec versions because it's just a warning.

Closes #131.

### DIFF
--- a/base/util/file.go
+++ b/base/util/file.go
@@ -35,3 +35,112 @@ func EnsurePathWithinFilesDir(path, filesDir string) error {
 	}
 	return nil
 }
+
+/// CheckForDecimalMode fails if the specified mode appears to have been
+/// incorrectly specified in decimal instead of octal.
+func CheckForDecimalMode(mode int, directory bool) error {
+	correctedMode, ok := decimalModeToOctal(mode)
+	if !ok {
+		return nil
+	}
+	if !isTypicalMode(mode, directory) && isTypicalMode(correctedMode, directory) {
+		return common.ErrDecimalMode
+	}
+	return nil
+}
+
+/// isTypicalMode returns true if the specified mode is unsurprising.
+/// It returns false for some modes that are unusual but valid in limited
+/// cases.
+func isTypicalMode(mode int, directory bool) bool {
+	// no permissions is always reasonable (root ignores mode bits)
+	if mode == 0 {
+		return true
+	}
+
+	// test user/group/other in reverse order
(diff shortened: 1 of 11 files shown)


jlebon:
This seems reasonable to me. To confirm, if the user really did mean that mode, then to squash this warning, they'd just have to convert it to octal (and make sure to add a leading zero), right?

jlebon:
Hmm, it's tempting to just warn on all non-octal modes instead of special-casing some possibly mistypes. Is that too strong-handed?

arithx (approved):

bgilbert:
Rebased and updated to use `testify`.

arithx (approved):

PR #158, opened 2020-11-23
base/util: add function to merge configs and TranslationSets

bgilbert:
We can use this to implement desugaring by generating a separate desugared config and merging it with the translated one.

### DIFF
--- a/base/util/merge.go
+++ b/base/util/merge.go
@@ -1,0 +1,61 @@
+// Copyright 2020 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.)
+
+package util
+
+import (
+	"fmt"
+
+	"github.com/coreos/fcct/translate"
+
+	"github.com/coreos/ignition/v2/config/merge"
+)
+
+// MergeTranslatedConfigs merges a parent and child config and returns the
+// result.  It also generates and returns the merged TranslationSet by
+// mapping the parent/child TranslationSets through the merge transcript.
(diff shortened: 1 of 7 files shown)


base/util/merge_test.go
```diff
@@ -0,0 +1,156 @@
...
+				{path.New("in", "bad", 1), path.New("out", "systemd", "units", 0, "name")},
+				// parent field overridden in child, should be clobbered
+				{path.New("in", "bad", 2), path.New("out", "systemd", "units", 0, "contents")},
+				// parent field not overridden in child
+				{path.New("in", "good", 1), path.New("out", "systemd", "units", 0, "enabled")},
+				// parent key not specified in child
+				{path.New("in", "good", 2), path.New("out", "systemd", "units", 1, "name")},
+				// parent field not specified in child
+				{path.New("in", "good", 3), path.New("out", "systemd", "units", 1, "contents")},
+				// other fields omitted from translation set
```
arithx, line 74: Is the idea behind omitting the `TranslationSet` for `elephant.service` to test that we can handle cases where there's a field in the result config / transcription that we don't have a corresponding `TranslationSet` for?

base/util/merge.go
```diff
@@ -0,0 +1,61 @@
...
+func MergeTranslatedConfigs(parent interface{}, parentTranslations translate.TranslationSet, child interface{}, childTranslations translate.TranslationSet) (interface{}, translate.TranslationSet) {
+	// mappings:
+	//   left:  parent or child translate.TranslationSet
+	//   right: merge.Transcript
+
+	// merge configs
+	result, right := merge.MergeStructTranscribe(parent, child)
+
+	// merge left and right mappings into new TranslationSet
+	if parentTranslations.FromTag != childTranslations.FromTag || parentTranslations.ToTag != childTranslations.ToTag {
```
arithx, line 37: Does this prevent us from being able to potentially add merged FCCT configs in the future?

arithx (approved):
Two questions, generally looks good.

base/util/merge_test.go
bgilbert: Yes.  The right side includes mappings for some fields and intermediate structs which might not be listed on the left side.  For example, `$.ignition.version` doesn't get a translation from `base`.

base/util/merge.go
bgilbert: It shouldn't.  If we merge FCCs after translating them, both parent and child will have TranslationSets mapping `yaml` -> `json`, as they do now.  If we merge FCCs before translating them, this function won't be useful anyway, since it assumes there's already been one level of translation before the merge occurs.

bgilbert:
Rebased, and updated the vendoring.

PR #162, opened 2020-11-24
Add boot_device sugar

bgilbert:
Add sugar for mirroring the root disk (all partitions) and/or encrypting the root partition.  Design document in https://github.com/coreos/enhancements/pull/3.

Example FCC:

```yaml
variant: fcos
version: 1.3.0-experimental
boot_device:
  # Specifies a disk layout template.  Supported values are aarch64,
  # ppc64le, x86_64 (default).
  layout: x86_64
  mirror:
    # If specified, mirror every default partition.  Two or more devices
    # are required.
    devices:
      - /dev/vda
      - /dev/vdb
  luks:
    # If tang and/or tpm2 is specified, encrypt the root partition.
    tang:
      - url: https://example.com/
        thumbprint: x
    tpm2: true
    threshold: 2
```

### DIFF
--- a/config/fcos/v1_3_exp/translate.go
+++ b/config/fcos/v1_3_exp/translate.go
@@ -15,12 +15,57 @@
 package v1_3_exp
 
 import (
+	"fmt"
+
+	baseutil "github.com/coreos/fcct/base/util"
 	"github.com/coreos/fcct/config/common"
 	cutil "github.com/coreos/fcct/config/util"
+	"github.com/coreos/fcct/translate"
 
+	"github.com/coreos/ignition/v2/config/util"
 	"github.com/coreos/ignition/v2/config/v3_2/types"
+	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
+
+const (
+	biosTypeGuid = "21686148-6449-6E6F-744E-656564454649"
+	prepTypeGuid = "9E1A2D38-C612-4316-AA26-8B49521E5A8B"
+	espTypeGuid  = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+
+	// The partition layout implemented in this file replicates
+	// the layout of the OS image defined in:
+	// https://github.com/coreos/coreos-assembler/blob/master/src/create_disk.sh
+	//
+	// Exception: we don't try to skip unused partition numbers,
+	// because specifying a partition number would prevent child
+	// configs from overriding partition fields using the partition
+	// label as the lookup key.
+	//
+	// It's not critical that we match that layout exactly; the hard
+	// constraints are:
+	//   - The desugared partition cannot be smaller than the one it
+	//     replicates
+	//   - The new BIOS-BOOT partition (and maybe the PReP one?) must be
+	//     at the same offset as the original
+	//
+	// Do not change these constants!  New partition layouts must be
+	// encoded into new layout templates.
+	biosV1SizeMiB = 1
+	prepV1SizeMiB = 4
+	espV1SizeMiB  = 127
+	bootV1SizeMiB = 384
+)
+
+// ToIgn3_2Unvalidated translates the config to an Ignition config.  It also
+// returns the set of translations it did so paths in the resultant config
+// can be tracked back to their source in the source config.  No config
+// validation is performed on input or output.
+func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Config, translate.TranslationSet, report.Report) {
+	ret, ts, r := c.Config.ToIgn3_2Unvalidated(options)
+	r.Merge(c.processBootDevice(&ret, &ts, options))
+	return ret, ts, r
+}
 
 // ToIgn3_2 translates the config to an Ignition config.  It returns a
 // report of any errors or warnings in the source and resultant config.  If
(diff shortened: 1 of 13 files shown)


arithx:
The current state of the code LGTM

bgilbert:
Ready for review!

config/fcos/v1_3_exp/translate.go
```diff
@@ -37,3 +53,173 @@ func (c Config) ToIgn3_2(options common.TranslateOptions) (types.Config, report.
...
+		rootDevice = "/dev/md/md-root"
+	default:
+		panic("can't happen")
+	}
+	rootFilesystem := types.Filesystem{
+		Device:         rootDevice,
+		Format:         util.StrToPtr("xfs"),
+		Label:          util.StrToPtr("root"),
+		WipeFilesystem: util.BoolToPtr(true),
+	}
```
jlebon, line 208: This just made me realize something. In cosa, [we enable reflinks](https://github.com/coreos/coreos-assembler/blob/7bf455a17d7033facbf6207c8d93557415f64aee/src/create_disk.sh#L239). In tests and documentation that reformat the rootfs to XFS, we're not re-adding it. Ideally, there'd be a way to just "copy" the settings from the rootfs we nuked, but short of that we should include it in our docs and make sure we test it too.

Anyway, circling back to here, since it's sugar, I think we should keep carry our opinion forward of enabling reflinks to match the original configuration.

docs/configuration-v1_3-exp.md
```diff
@@ -190,6 +190,16 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_gid_** (integer): the group ID of the new group.
     * **_password_hash_** (string): the hashed password of the new group.
     * **_system_** (bool): whether or not the group should be a system group. This only has an effect if the group doesn't exist yet.
+* **_boot_device_** (object): describes the desired boot device configuration.
+  * **_layout_** (string): the disk layout of the target OS image. This is only needed if also specifying `luks` or `mirror`. Supported values are `aarch64`, `ppc64le`, and `x86_64`. Defaults to `x86_64`.
+  * **_luks_** (object): describes the clevis configuration for encrypting the root filesystem.
```
jlebon, line 195: I think we discussed this, but just to check: are you purposely ignoring `custom` here?

config/fcos/v1_3_exp/translate_test.go
```diff
@@ -0,0 +1,1153 @@
...
+)
+
+// Most of this is covered by the Ignition translator generic tests, so just test the custom bits
+
+// TestTranslateBootDevice tests translating the FCC boot_device section.
+func TestTranslateBootDevice(t *testing.T) {
+	tests := []struct {
+		in         Config
+		out        types.Config
+		exceptions []translate.Translation
```
jlebon, line 39: Hmm, should this be called e.g. `translations` instead?

config/fcos/v1_3_exp/translate_test.go
```diff
@@ -0,0 +1,1153 @@
...
+				{path.New("yaml", "boot_device", "mirror"), path.New("json", "storage", "filesystems", 1, "format")},
+				{path.New("yaml", "boot_device", "mirror"), path.New("json", "storage", "filesystems", 1, "label")},
+				{path.New("yaml", "boot_device", "mirror"), path.New("json", "storage", "filesystems", 1, "wipeFilesystem")},
+				{path.New("yaml", "boot_device"), path.New("json", "storage", "filesystems", 2, "device")},
+				{path.New("yaml", "boot_device"), path.New("json", "storage", "filesystems", 2, "format")},
+				{path.New("yaml", "boot_device"), path.New("json", "storage", "filesystems", 2, "label")},
+				{path.New("yaml", "boot_device"), path.New("json", "storage", "filesystems", 2, "wipeFilesystem")},
+			},
+		},
+		// 2-disk mirror + LUKS with overridden root partition size
```
jlebon, line 917: On the subject of overridding the root partition size and just general overrides, are we missing docs somewhere about what e.g. labels to use to make sure it matches up with what the sugar outputs?

config/fcos/v1_3_exp/translate_test.go
bgilbert: No, this is the `exceptions` argument to `VerifyTranslations()`.  It just so happens that every translation here is an exception (except for two of them, oops).

config/fcos/v1_3_exp/translate.go
```diff
@@ -37,3 +53,173 @@ func (c Config) ToIgn3_2(options common.TranslateOptions) (types.Config, report.
...
+				Label:    util.StrToPtr(fmt.Sprintf("esp-%d", labelIndex)),
+				SizeMiB:  util.IntToPtr(127),
+				TypeGUID: util.StrToPtr("C12A7328-F81F-11D2-BA4B-00A0C93EC93B"),
+			}, types.Partition{
+				Label:   util.StrToPtr(fmt.Sprintf("boot-%d", labelIndex)),
+				SizeMiB: util.IntToPtr(384),
+			}, types.Partition{
+				Label: util.StrToPtr(fmt.Sprintf("root-%d", labelIndex)),
+			})
+			renderedTranslations.AddFromCommonSource(path.New("yaml", "boot_device", "mirror", "devices", i), path.New("json", "storage", "disks", len(rendered.Storage.Disks)), disk)
```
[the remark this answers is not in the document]
bgilbert: Yeah, it needs to do that to provide input line/column numbers corresponding to JSON validation errors.

config/fcos/v1_3_exp/translate.go
```diff
@@ -37,3 +53,173 @@ func (c Config) ToIgn3_2(options common.TranslateOptions) (types.Config, report.
...
+					TypeGUID: util.StrToPtr("9E1A2D38-C612-4316-AA26-8B49521E5A8B"),
+				})
+			}
+			disk.Partitions = append(disk.Partitions, types.Partition{
+				Label:    util.StrToPtr(fmt.Sprintf("esp-%d", labelIndex)),
+				SizeMiB:  util.IntToPtr(127),
+				TypeGUID: util.StrToPtr("C12A7328-F81F-11D2-BA4B-00A0C93EC93B"),
+			}, types.Partition{
+				Label:   util.StrToPtr(fmt.Sprintf("boot-%d", labelIndex)),
+				SizeMiB: util.IntToPtr(384),
```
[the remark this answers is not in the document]
bgilbert: Constants are a good idea.  The source of truth isn't `create_disk.sh`, though, it's external standards.  For example, the ESP GUID is defined in the UEFI spec.

docs/configuration-v1_3-exp.md
bgilbert: Yes.  Use of the sugar is not mandatory, and users who want to deeply customize their LUKS (or RAID) configuration can write the low-level config stanzas directly.

docs/configuration-v1_3-exp.md
```diff
@@ -190,6 +190,16 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_gid_** (integer): the group ID of the new group.
     * **_password_hash_** (string): the hashed password of the new group.
     * **_system_** (bool): whether or not the group should be a system group. This only has an effect if the group doesn't exist yet.
+* **_boot_device_** (object): describes the desired boot device configuration.
+  * **_layout_** (string): the disk layout of the target OS image. This is only needed if also specifying `luks` or `mirror`. Supported values are `aarch64`, `ppc64le`, and `x86_64`. Defaults to `x86_64`.
```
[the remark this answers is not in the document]
bgilbert: > That makes it seems like there's a reason for using boot_device without either of those, but there isn't, right?

That's exactly the problem I'm trying to avoid: users thinking they need to specify the `layout` if they're not specifying `luks` or `mirror`.  Your approach seems better for that.  (There's not really a way to validate it, though, since we don't use struct pointers generally.)

config/fcos/v1_3_exp/translate.go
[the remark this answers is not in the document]
bgilbert: Oh, you meant the source of truth for the partition layout.  Sure, makes sense.

config/fcos/v1_3_exp/translate_test.go
bgilbert: Yeah, we need docs for this feature generally.  Let's defer that to a followup?

config/fcos/v1_3_exp/translate.go
bgilbert: I agree that we should replicate any unusual FS options into the desugared config.  `mkfs.xfs(8)` on both Fedora 33 and CentOS 8 says that reflinks are enabled by default, so I don't think we need to do anything here?

bgilbert:
Updated!

config/fcos/v1_3_exp/translate.go
jlebon: Ahh nice, didn't realize that was the default now. That means we should be able to drop it from cosa too. Will add that to the list.

jlebon (approved):
LGTM!

PR #178, opened 2020-12-20
config/fcos/v1_[34]: un-RAID ESP

bgilbert:
It's not safe to RAID the ESP because the firmware might decide to write to an individual replica, causing the RAID to desynchronize.  Instead, create a separate filesystem for each replica.

Don't address backward compatibility because we're making this change soon after the functionality landed, and before it was documented in the OS.

Part of https://github.com/coreos/fedora-coreos-config/pull/794.  Fixes [RHBZ 1909453](https://bugzilla.redhat.com/show_bug.cgi?id=1909453).  Discussion in https://github.com/coreos/fedora-coreos-config/pull/718 and design doc update in https://github.com/coreos/enhancements/pull/4.


arithx (approved):

sohankunkerkar (approved):
LGTM

bgilbert:
Holding until https://github.com/coreos/fedora-coreos-config/pull/794 lands.

PR #180, opened 2020-12-23
docs: Update fcct download instructions

travier:
This places the container option first, Fedora package second and the binaries last, similarly to how this is organised in FCOS tutorial docs: https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-setup/

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -11,22 +11,13 @@ Fedora CoreOS Configs are YAML files conforming to `fcct`'s schema. For more inf
 
 ### Getting FCCT
 
-`fcct` can be downloaded as a standalone binary or run as a container with docker or podman.
+`fcct` can be run from a container image with `podman` or `docker`, installed from Fedora package repositories or downloaded as a standalone binary.
 
-#### Standalone binary
+Using the official container images is the recommended option.
 
-Download the latest version of `fcct` and the detached signature from the [releases page](https://github.com/coreos/fcct/releases). Verify it with gpg:
+#### Container image
 
-```
-gpg --verify <detached sig> <fcct binary>
-```
-You may need to download the [Fedora signing keys](https://getfedora.org/static/fedora.gpg) and import them with `gpg --import <key>` if you have not already done so.
-
-New releases of `fcct` are backwards compatible with old releases unless otherwise noted.
-
-#### Container
-
-This example uses podman, but docker can also be used.
+This example uses `podman`, but `docker` can also be used.
 
 ```bash
 # Pull the latest release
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -39,6 +30,38 @@ podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.f
 podman run --rm -v /path/to/your_config.fcc:/config.fcc:z quay.io/coreos/fcct:release --pretty --strict /config.fcc > transpiled_config.ign
 ```
 
+You may also setup the following alias in your shell configuration:
```
bgilbert, line 33: `setup` is a noun; the verb is`set up`.  But `create` might be better.

docs/getting-started.md
```diff
@@ -39,6 +30,38 @@ podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.f
...
+```
+alias fcct='podman run --rm --tty --interactive \
+            --security-opt label=disable        \
+            --volume ${PWD}:/pwd --workdir /pwd \
+            quay.io/coreos/fcct:release'
+```
+
+#### Distribution packages
+
+`fcct` is available from the Fedora packages repositories:
```
bgilbert, line 44: `package repositories`

docs/getting-started.md
```diff
@@ -39,6 +30,38 @@ podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.f
...
+
+```
+$ sudo dnf install -y fcct
+```
+
+Although this should work, we recommend that you run fcct from the container (as described in the previous section) to make sure that you always use the latest version. If you use the packaged version, make sure that your system is fully updated to the latest Fedora stable release.
```
bgilbert, line 50: I think this is probably overstated.  We push new fcct releases to all supported Fedora releases, and there aren't generally adverse consequences from using older versions.  The main consequence is that newer specs may not be supported.

docs/getting-started.md
```diff
@@ -39,6 +30,38 @@ podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.f
...
+Download the latest version of `fcct` and the detached signature from the [releases page](https://github.com/coreos/fcct/releases). Verify it with gpg:
+
+```
+gpg --verify <detached sig> <fcct binary>
+```
+You may need to download the [Fedora signing keys](https://getfedora.org/static/fedora.gpg) and import them with `gpg --import <key>` if you have not already done so.
+
+New releases of `fcct` are backwards compatible with old releases unless otherwise noted.
+
+Although this should work, we recommend that you run fcct from the container (as described in the first section) to make sure that you always use the latest version.
```
bgilbert, line 63: Using the standalone binary is somewhat worse than the distro package, but I'd still be inclined to drop this line.

docs/getting-started.md
travier: Went with 'add'

bgilbert (approved):

PR #186, opened 2020-12-24
Stop building non-Linux-x86_64 binaries during container build

bgilbert:
We don't use them in the resulting container.

Also rename `build_releases` to clarify that it's only used for the container build now.

Fixes https://github.com/coreos/fcct/issues/151.


sohankunkerkar (approved):
LGTM

arithx (approved):

PR #190, opened 2021-01-06
NEWS: add news for 0.9.0

sohankunkerkar:


NEWS
```diff
@@ -1,3 +1,21 @@
+2021-01-05 FCCT 0.9.0
+
+  Bug Fixes:
+
+    - Avoid ESP RAID desynchronization by creating independent ESP filesystems
+  
+  Misc Changes:
+
+    - Avoid building non-Linux-x86_64 binaries
```
bgilbert, line 9: Let's drop this.  The change only affects the build script for the container image, which was shipping only the Linux x86_64 binary anyway.  IOW, the change isn't user-visible.

bgilbert (commented):
LGTM generally.

bgilbert (approved):

PR #203, opened 2021-02-11
docs: fix `resize` example

jlebon:
The example there doesn't work as one would expect: it actually creates
a new partition labeled `root` with the given size. We need to pin it
using the partition number to make sure it affects the existing one.

### DIFF
--- a/docs/migrating-configs.md
+++ b/docs/migrating-configs.md
@@ -58,7 +58,8 @@ storage:
   disks:
     - device: /dev/sda
       partitions:
-        - label: root
+        - number: 4
+          label: root
           size_mib: 16384
           resize: true
 ```


docs/migrating-configs.md
```diff
@@ -58,7 +58,8 @@ storage:
   disks:
     - device: /dev/sda
       partitions:
-        - label: root
+        - number: 4
+          label: root
```
bgilbert, line 62: For clarity, should we drop the `label` attribute?

jlebon: I like it as a safety-check so that it ensures that you're using the right now. WDYT?

bgilbert (approved):

PR #206, opened 2021-02-23
Add TranslationSet entries for various interior nodes

bgilbert:
We've mostly assumed that TranslationSets don't need to include interior nodes.  However, some config validation can already generate report entries pointing to interior nodes, and there's no reason that shouldn't happen in general, so we need a TranslationSet that's as comprehensive as possible.  This will be especially true once we reparent the output config within another struct, since in that case the automatic fallback to a 1:1 mapping from input to output paths will no longer be useful.

This PR fixes many of the cases, but there's some followup work remaining:
- Ignition merge transcripts don't report the sources of structs which are combinations of parent and child, and we can't re-derive that information in the caller.  Ignition should be able to report the parent and child paths for a merged struct, but that requires an API change.
- Call `DebugVerifyCoverage()` in more tests.
- Consider dropping the automatic 1:1 fallback.


bgilbert:
Updated to vendor https://github.com/coreos/ignition/pull/1166.  Ready for review.

arithx (approved):

PR #212, opened 2021-03-10
Add openshift/4.8.0-exp that produces MachineConfigs by default

bgilbert:
Rename `rhcos`/`0.2.0-experimental` to `openshift`/`4.8.0-experimental`.  Use `openshift` as the variant name instead of `rhcos` because RHCOS is an implementation detail of OpenShift, and also would leave OKD in an awkward position.  By reusing OpenShift version numbers, we're setting ourselves up to eventually violate semver, but it doesn't seem reasonable to create a novel set of user-facing version numbers that isn't connected to anything else.

Have the new spec generate MachineConfigs by default, but also accept a new `-r/--raw/Raw` option to generate an Ignition config.  This allows FCCT to be used for sugar (e.g. for [boot disk mirroring](https://docs.openshift.com/container-platform/4.7/installing/install_config/installing-customizing.html#installation-special-config-mirrored-disk_installing-customizing)) and validation (e.g. rejecting users other than `core`) without modifying the pointer config, and without requiring MCO changes.  We'll have a new spec version for each OpenShift release, so we'll also have the ability to change the output format as needed by future MCO releases.

As an implementation detail, ship a simplified copy of the [MachineConfig structs](https://github.com/openshift/machine-config-operator/blob/82868e63176fee2bc806c1deb308ed1fc8965d84/pkg/apis/machineconfiguration.openshift.io/v1/types.go) so we don't have to pull in a bunch of k8s code as dependencies.  (And to avoid a circular dependency with the MCO.)

Sample `input.occ`:

```yaml
variant: openshift
version: 4.8.0-experimental
metadata:
  name: create-etc-sample
  labels:
storage:
  files:
    - path: /etc/sample
      contents:
        inline: hello world
```

`fcct input.occ` produces:

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
  name: create-etc-sample
spec:
  config:
    ignition:
      version: 3.2.0
    storage:
      files:
      - contents:
          source: data:,hello%20world
        path: /etc/sample
```

And `fcct --pretty --raw input.occ` produces:

```json
{
  "ignition": {
    "version": "3.2.0"
  },
  "storage": {
    "files": [
      {
        "path": "/etc/sample",
        "contents": {
          "source": "data:,hello%20world"
        }
      }
    ]
  }
}
```

@yuqi-zhang, @sinnykumari, and/or other folks from the MCO team, wdyt?

### DIFF
--- a/docs/config-openshift-v4_8-exp.md
+++ b/docs/config-openshift-v4_8-exp.md
@@ -13,6 +13,9 @@ The OpenShift configuration is a YAML document conforming to the following speci
 
 * **variant** (string): used to differentiate configs for different operating systems. Must be `rhcos` for this specification.
 * **version** (string): the semantic version of the spec for this document. This document is for version `4.8.0-experimental` and generates Ignition configs with version `3.2.0`.
+* **metadata** (object): metadata about the generated MachineConfig resource. Respected when rendering to a MachineConfig, ignored when rendering directly to an Ignition config.
+  * **name** (string): a unique [name][k8s-names] for this MachineConfig resource.
+  * **_labels_** (object): string key/value pairs to apply as [Kubernetes labels][k8s-labels] to this MachineConfig resource.
 * **ignition** (object): metadata about the configuration itself.
   * **_config_** (objects): options related to the configuration.
     * **_merge_** (list of objects): a list of the configs to be merged to the current config.
@@ -202,6 +205,8 @@ The OpenShift configuration is a YAML document conforming to the following speci
   * **_mirror_** (object): describes mirroring of the boot disk for fault tolerance.
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 
+[k8s-names]: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+[k8s-labels]: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
 [part-types]: http://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs
 [rfc2397]: https://tools.ietf.org/html/rfc2397
 [systemd-escape]: https://www.freedesktop.org/software/systemd/man/systemd-escape.html
--- a/docs/config-openshift-v4_8-exp.md
+++ b/docs/config-openshift-v4_8-exp.md
@@ -1,18 +1,18 @@
 ---
 layout: default
-title: RHEL CoreOS v0.2.0-experimental
+title: OpenShift v4.8.0-experimental
 parent: Configuration specifications
 nav_order: 100
 ---
 
-# RHEL CoreOS Specification v0.2.0-experimental
+# OpenShift Specification v4.8.0-experimental
 
 **Note: This configuration is experimental and has not been stabilized. It is subject to change without warning or announcement.**
 
-The RHEL CoreOS configuration is a YAML document conforming to the following specification, with **_italicized_** entries being optional:
+The OpenShift configuration is a YAML document conforming to the following specification, with **_italicized_** entries being optional:
 
 * **variant** (string): used to differentiate configs for different operating systems. Must be `rhcos` for this specification.
-* **version** (string): the semantic version of the spec for this document. This document is for version `0.2.0-experimental` and generates Ignition configs with version `3.3.0-experimental`.
+* **version** (string): the semantic version of the spec for this document. This document is for version `4.8.0-experimental` and generates Ignition configs with version `3.3.0-experimental`.
 * **ignition** (object): metadata about the configuration itself.
   * **_config_** (objects): options related to the configuration.
     * **_merge_** (list of objects): a list of the configs to be merged to the current config.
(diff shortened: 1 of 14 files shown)


docs/config-openshift-v4_8-exp.md
```diff
@@ -1,18 +1,18 @@
 ---
 layout: default
-title: RHEL CoreOS v0.2.0-experimental
+title: OpenShift v4.8.0-experimental
 parent: Configuration specifications
 nav_order: 100
 ---
 
-# RHEL CoreOS Specification v0.2.0-experimental
+# OpenShift Specification v0.2.0-experimental
```
arithx, line 8:
```suggestion
# OpenShift Specification v4.8.0-experimental
```

sinnykumari:
I don't have full context here, is the usecase of this is that user can write openshift coreos config (yaml config) and can use fcct tool to generate MachineConfig?

cc @kikisdeliveryservice

bgilbert:
@sinnykumari Yup, that's right.  For example, it would let us convert the [mirrored disk instructions](https://docs.openshift.com/container-platform/4.7/installing/install_config/installing-customizing.html#installation-special-config-mirrored-disk_installing-customizing) to use a MachineConfig manifest.  Other workflows in the customization docs could also be switched to preprocess through fcct in cases where its sugar and validation capabilities would be useful.

yuqi-zhang:
In the MCO today I believe we use FCCT to render our templates into machineconfigs, although we make no explicit reference to the schema type. Is that something we should change? Or is that perhaps unrelated to this change

sinnykumari:
> @sinnykumari Yup, that's right. For example, it would let us convert the [mirrored disk instructions](https://docs.openshift.com/container-platform/4.7/installing/install_config/installing-customizing.html#installation-special-config-mirrored-disk_installing-customizing) to use a MachineConfig manifest.

+1

> Other workflows in the customization docs could also be switched to preprocess through fcct in cases where its sugar and validation capabilities would be useful.

agree, for better user experience it would be nice to be able to generate rest of MachineConfig file fields as well.
It makes me think renaming the tool from `fcct` to more generic would be useful as this will get used by FCOS, OKD and OCP.

And current .occ format looks fine to me.

LorbusChris:
+1 from my side as well.

Just to xref some previous ideas in this area:
- https://github.com/openshift/enhancements/pull/525
- https://github.com/openshift/machine-config-operator/pull/1980
- And also @cgwalters' suggestion of creating a `coreosconfig` CRD as replacement for MCs-before-rendering: https://github.com/openshift/enhancements/pull/525#issuecomment-736814610

Do I see this correctly and this RFC will be able tp obsolete these approaches?

bgilbert:
No major objections raised so far; dropping `RFC`!

bgilbert:
This part of the new FCC syntax is worth some thought:

```yaml
metadata:
  labels:
```

Nowhere else in the FCC or Ignition specs do we allow user-specified map keys.  Normally we'd implement this functionality like so:

```yaml
metadata:
  labels:
    - key: machineconfiguration.openshift.io/role
      value: worker
```

I'm concerned that if we use a datatype here that we don't use elsewhere, and don't plan to use elsewhere, that it will cause complications.  It already introduces a presentation issue in [documentation](https://github.com/coreos/fcct/pull/212/files#diff-29434339d33f230bf3fd1d2e7f2631b19093237bb9f4c1c7aa7b0fa6dc028770R19) and requires some additional code in `translate/util.go`.  On the other hand, a more Ignition-like structure would be inconsistent with k8s norms.

Thoughts?

arithx:
I would lean towards the former example. I think making it clear that we're directly passing those arguments through into the resulting MachineConfig is sufficient (as well giving us the mentioned benefit of matching the k8s norms).

internal/main.go
```diff
@@ -45,6 +45,7 @@ func main() {
 	pflag.Lookup("debug").Hidden = true
 	pflag.BoolVarP(&options.Strict, "strict", "s", false, "fail on any warning")
 	pflag.BoolVarP(&options.Pretty, "pretty", "p", false, "output formatted json")
+	pflag.BoolVarP(&options.Raw, "raw", "r", false, "output only Ignition config, without any wrapper")
```
arithx, line 48: Do you think we should mention this is only valid on `openshift` variants or maybe throw a warning/error if we see it set in other variants? I'm not sure that `without any wrapper` means anything without context.

arithx (approved):
LGTM

jlebon:
> This part of the new FCC syntax is worth some thought:

Agreed with @arithx. Just matching the k8s semantics would be a nicer UX. Documentation-wise, feels like we can just leave out the awkward `arbitrary-key` line entirely and enhance the `labels` line to something like:

>   * **_labels_** (object): string key/value pairs to apply as [Kubernetes labels][k8s-labels] to this MachineConfig.

?

ashcrow (approved):

internal/main.go
bgilbert: Changed it to `never wrap in a MachineConfig; force Ignition output`.  For the other variants, we always output raw Ignition, so it feels odd to complain if the user redundantly requests it.

PR #214, opened 2021-03-19
openshift/v4_8_exp: support top-level MachineConfig fields

bgilbert:
Add an `openshift` section containing the fields at the top level of a MachineConfig:

```yaml
variant: openshift
version: 4.8.0-experimental
metadata:
  name: z
openshift:
  kernel_arguments: [a, b, c]
  extensions: [d, e]
  fips: true
  kernel_type: realtime
```

Output:

```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  name: z
spec:
  config:
    ignition:
      version: 3.2.0
  extensions:
  - d
  - e
  fips: true
  kernelArguments:
  - a
  - b
  - c
  kernelType: realtime
```

Report warnings in `--raw` mode if any such fields are specified, since we can't represent them directly in an Ignition config.

Followup to #212.

cc @sinnykumari

### DIFF
--- a/config/openshift/v4_8_exp/translate_test.go
+++ b/config/openshift/v4_8_exp/translate_test.go
@@ -1,0 +1,49 @@
+// Copyright 2021 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.)
+
+package v4_8_exp
+
+import (
+	"testing"
+
+	"github.com/coreos/fcct/config/common"
+
+	"github.com/coreos/ignition/v2/config/util"
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestElidedFieldWarning tests that we warn when transpiling fields to an
+// Ignition config that can't be represented in an Ignition config.
+func TestElidedFieldWarning(t *testing.T) {
+	in := Config{
+		Metadata: Metadata{
+			Name: "z",
+		},
+		OpenShift: OpenShift{
+			KernelArguments: []string{"a", "b"},
+			FIPS:            util.BoolToPtr(true),
+			KernelType:      util.StrToPtr("realtime"),
+		},
+	}
+
+	var expected report.Report
+	expected.AddOnWarn(path.New("yaml", "openshift", "kernel_arguments"), common.ErrFieldElided)
+	expected.AddOnWarn(path.New("yaml", "openshift", "fips"), common.ErrFieldElided)
+	expected.AddOnWarn(path.New("yaml", "openshift", "kernel_type"), common.ErrFieldElided)
+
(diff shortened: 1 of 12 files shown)


config/openshift/v4_8_exp/translate_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/coreos/vcontext/report"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestElidedFieldWarning tests that we warn when transpiling fields to an
+// Ignition config that can't be represented in an Ignition config.
+func TestElidedFieldWarning(t *testing.T) {
+	in := Config{
+		Metadata: Metadata{
+			Name: "z",
```
arithx, line 32: It feels a bit awkward that `metadata.name` is a required field but is elided and not warned about when rendering raw Ignition. Although I'm not sure if we could get around the circular dependency issue if we wanted to check the status of `raw` so I'm okay just living with it.

config/util/util.go
```diff
@@ -160,6 +160,28 @@ func TranslateBytesYAML(input []byte, container interface{}, translateMethod str
 	return yamlCfg, r, err
 }
 
+// Report an ErrFieldElided warning for any non-zero top-level fields in the
+// specified output struct.  Append the field's `json` struct tag to the
+// specified path prefix, then map the result backward through the
+// TranslationSet to get the original input field path.
+func CheckForElidedFields(struct_ interface{}, c path.ContextPath, ts translate.TranslationSet) report.Report {
```
arithx, line 167: Should we add `metadata.labels` to this check?

arithx (approved):
Two questions, otherwise code LGTM

sinnykumari:
Overall this looks good, one small change.

Along with name, we also need to validate and add test for `labels` field of `metadata`. labels are used by machine-config-controller to know which pool this MachineConfig will be applied to.

Example valid MachineConfig.

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
  name: test-worker-mc
spec:
  config:
    ignition:
      version: 3.0.0
  extensions:
    - foo
  ```

arithx:
> Along with name, we also need to validate and add test for labels field of metadata. labels are used by machine-config-controller to know which pool this MachineConfig will be applied to.

So then `machineconfiguration.openshift.io/role` is a required key of `metadata.labels`? Does it have any limitations on the value?

sinnykumari:
> So then `machineconfiguration.openshift.io/role` is a required key of `metadata.labels`? Does it have any limitations on the value?

Since we allow creating custom machineconfigpool, value can be anything along with worker and master.

bgilbert:
Added validation for `machineconfiguration.openshift.io/role` in https://github.com/coreos/fcct/pull/215.

config/openshift/v4_8_exp/translate_test.go
bgilbert: I intentionally didn't warn on the `metadata` fields, on the theory that they don't affect how the machine is configured.  The `openshift` fields do affect the machine, so it could be surprising if those changes are not applied.

arithx: That's fair, although I couldn't say with certainty that there isn't something potentially keying off a metadata field that might trigger machine-level modifications.

bgilbert:
I want to refactor this to support some future additional validation.

config/openshift/v4_8_exp/translate_test.go
bgilbert: I can't either.  But if we warn on `metadata`, the `--raw` flag will always emit warnings and most of the time I suspect they wouldn't be meaningful.  So there's a risk of warning fatigue.

bgilbert:
Updated.

arithx (approved):

PR #218, opened 2021-03-20
openshift/v4_8_exp: set cipher for LUKS volumes if FIPS enabled

### LINKED ISSUE #198: Consider sugar for FIPS mode
LUKS with FIPS 140-2 (not 140-3) requires `options: [--cipher, aes-cbc-essiv:sha256]`.  As a UX improvement, we could add a `fips: true` field to the `luks` section, but it probably doesn't make sense to enable FIPS compatibility on a per-volume basis.  @arithx suggested adding a global `fips` field which enables FIPS compatibility wherever possible.  For now, that flag could configure the cipher on any LUKS volume where a cipher isn't already set.

bgilbert:
If `openshift.fips` is specified, add `--cipher aes-cbc-essiv:sha256` to `storage.luks.*.options` unless a cipher is already specified in the options list.  The default cipher uses XTS, which isn't allowed by FIPS 140-2.

Fixes #198.  We key off of the `openshift.fips` field, rather than creating a new field shared with FCOS, since the former already exists and FCOS isn't likely to support FIPS.

### DIFF
--- a/config/openshift/v4_8_exp/translate.go
+++ b/config/openshift/v4_8_exp/translate.go
@@ -15,14 +15,24 @@
 package v4_8_exp
 
 import (
+	"strings"
+
 	"github.com/coreos/fcct/config/common"
 	"github.com/coreos/fcct/config/openshift/v4_8_exp/result"
 	cutil "github.com/coreos/fcct/config/util"
 	"github.com/coreos/fcct/translate"
 
+	"github.com/coreos/ignition/v2/config/util"
 	"github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
+)
+
+const (
+	// FIPS 140-2 doesn't allow the default XTS mode
+	fipsCipherOption      = types.LuksOption("--cipher")
+	fipsCipherShortOption = types.LuksOption("-c")
+	fipsCipherArgument    = types.LuksOption("aes-cbc-essiv:sha256")
 )
 
 // ToMachineConfig4_8Unvalidated translates the config to a MachineConfig.  It also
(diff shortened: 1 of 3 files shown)


config/openshift/v4_8_exp/translate.go
```diff
@@ -121,3 +131,31 @@ func ToConfigBytes(input []byte, options common.TranslateBytesOptions) ([]byte,
...
+	if !util.IsTrue(mc.Spec.FIPS) {
+		return ts
+	}
+
+OUTER:
+	for i := range mc.Spec.Config.Storage.Luks {
+		luks := &mc.Spec.Config.Storage.Luks[i]
+		// Only add options if the user hasn't already specified
+		// fipsCipherOption.  Do this in-place, since config merging
+		// doesn't support conditional logic.
```
arithx, line 146: Might be more clear to reword the first sentence to something like `Only add options if the user hasn't already specified a cipher option.` to more clearly spell out that we're bailing on any cipher being present as the constant seems like it's FIPS related unless you go look up the value.

config/openshift/v4_8_exp/translate.go
```diff
@@ -121,3 +131,31 @@ func ToConfigBytes(input []byte, options common.TranslateBytesOptions) ([]byte,
...
+		for _, option := range luks.Options {
+			if option == fipsCipherOption {
```
arithx, line 148: Should this be changed out to `strings.HasPrefix(option, fipsCipherOption)`? I think `--cipher=my-cipher-here` should be valid.

bgilbert: Whoops, good catch.

bgilbert:
Updated.

arithx (approved):

PR #219, opened 2021-03-23
openshift/v4_8_exp: error on fields rejected or discouraged by MCO

bgilbert:
Error on fields that are rejected outright by the MCO, or that are unsupported by the MCO and we want to discourage.

@yuqi-zhang @sinnykumari @kikisdeliveryservice, WDYT?  Are there additional validations we should add?


config/common/errors.go
```diff
@@ -52,4 +52,5 @@ var (
 	ErrNameRequired      = errors.New("metadata.name is required")
 	ErrRoleRequired      = errors.New("machineconfiguration.openshift.io/role label is required")
 	ErrInvalidKernelType = errors.New("must be empty, \"default\", or \"realtime\"")
+	ErrMCOUnsupported    = errors.New("Machine Config Operator will not manage this resource after Ignition configures it")
```
darkmuggle, line 55:
```suggestion
	ErrMCOUnsupported    = errors.New("machine Config Operator will not manage this resource after Ignition configures it")
```

darkmuggle (commented):
I'm not intimately familiar with the deep magic going on here, although if I am following the code correctly, this is mostly dead-on. 

Given that I haven't worked on this code, I'll defer to another First Bootineer to review. The GoLang looks good, other than a stylistic nit.

darkmuggle:
FWIW, I *think* that a hard nack is better than a warning in the case of MCO parsing. My reasoning is that warning may be ignored, while a hard nack will educate the customer.

config/common/errors.go
bgilbert: The idea here was to keep the capitalization of the proper noun.

kikisdeliveryservice:
> FWIW, I _think_ that a hard nack is better than a warning in the case of MCO parsing. My reasoning is that warning may be ignored, while a hard nack will educate the customer.

@darkmug Just to clarify, would that then mean that they can't do it anyway? I did have a similar thought that they won't really pay attn to a warning but we still do have to let them make the change. I guess the Q is what's the happy medium

kikisdeliveryservice:
Overall, I really like the idea of giving users important info before they go forth into the world and make changes so I'm +1 on this idea (and think we could do more of it across the board).  Guess the question is how strong of a warning etc...  do we give them?

yuqi-zhang:
To clarify, the main use case for the openshift-4.8 schema would be translating it to either a MachineConfig to be provided during install time, or directly as ignition config to append to your node-bootstrap configs, right? In that case I think a warning is probably ok.

One other thing I can think of right now is that files/append is not handled, and will actually cause issues with MCO when the MCD takes over. Maybe we should consider failing on something like that.

darkmuggle:
> > @darkmuggle Just to clarify, would that then mean that they can't do it anyway? I did have a similar thought that they won't really pay attn to a warning but we still do have to let them make the change. I guess the Q is what's the happy medium

The idea with Benjamin's change is that you can take an MCC and then have `fcct` generate Ignition JSON. The reason I think we should hard nack is that some subset of uses will the interface of `fcct` to check their MCC payloads or even do the installation (i.e. user needs specific $THING at first boot before MCO renders the configuration, or user wants to see that it works against a generic RHCOS node without affecting their cluster, etc). In other words, I see this `fcct` interface as a developer tool, and having a hard NO provides very useful information. 

Then you have the whole idea of consistency. If `fcct` is going to understand MCC's, if the MCO would hard nack, then having consistent behavior is valuable.

ashcrow (commented):
Great idea!

kikisdeliveryservice:
I like the idea of this a lot, but tend to think the hard-nack @darkmuggle suggested sends a clearer message. wdyt @bgilbert ?

yuqi-zhang:
Question: what are we proposing to hard-nack on? Is it:
1. Anything the MCO would not handle after the fact, if provided through ignition or
2. Anything that would cause the MCO to error

For example, if you provide "disks" such that ignition does partitioning on the first run, while the MCO doesn't handle it, so long as you don't change that sub-field after the fact, you won't cause any errors. Indeed we officially document how to do this via ignition. This would fall under 1.

Another example would be append that I described earlier, which would fall under 2, where ignition can correctly provide it but as soon as the MCO takes over it would error.

My alternative proposal would be we hard nack anything that falls under 2, and warn on anything that falls under 1, to provide more information for the developer. This would definitely be a bit harder, but I can try to help look through how the distinction is made, since the MCO isn't actually very consistent on this.

bgilbert:
> The idea with Benjamin's change is that you can take an MCC and then have fcct generate Ignition JSON.

@yuqi-zhang is correct in https://github.com/coreos/fcct/pull/219#issuecomment-805170682: the input is an FCC (naming is still up in the air) and the output is a MachineConfig or (if `--raw` is specified on the command line) an Ignition config.  The latter might be used in some UPI provisioning cases, and the former could be used anywhere a preprocessor for MachineConfigs would be useful.  See https://github.com/coreos/fcct/pull/212 for more.

> My alternative proposal would be we hard nack anything that falls under 2, and warn on anything that falls under 1, to provide more information for the developer. This would definitely be a bit harder, but I can try to help look through how each case is handled, since the MCO isn't actually very consistent on this.

I was originally thinking case 1 would be split into two subcases: 1a) storage layout changes (basically, configuration handled by the Ignition disks stage), and 1b) filesystem layout changes, including user/group changes (handled by files stage).

For case 2, we should certainly hard nack (in FCCT we'd say "error") rather than warning; that's as easy as `s/AddOnWarn/AddOnError/`.  For 1a I assume we shouldn't emit any messages.  For 1b the current code warns but does not error (modulo any cases I've misclassified).  I'm okay with making 1b an error instead, depending how much we want to discourage the behavior.  If the user might intentionally want to do the thing (maybe there's a kbase somewhere that tells them to) and we error out, we're just encouraging them to bypass FCCT and hand-edit their configs.

@yuqi-zhang, if you'd be willing to look through the MCO code a bit, I can add whatever validation logic makes sense here.  I've pushed a change to error on `files[].append`.

arithx:
Code itself LGTM; As far as error/warn my line of thinking is pretty in line with @bgilbert's latest comment.

sinnykumari:
MCO does ignition config related validation at https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/update.go#L739 , today node would go degraded if unsupported ignition field has been updated. User will have to manually go and fix them. I also think that hard nack is better for clear messaging.

yuqi-zhang:
I feel like we should target this towards day 1 operation. Today we don't really have a good converter there I feel, which is maybe a gap this can help bridge for developers.

With that in mind and cross referencing some Openshift docs, I think `Storage.Directories` and `Storage.Links` and `Storage.Raid` is not referenced to performing any configuration Day 1, but wouldn't error if you did. We can upgrade these to Errors as well since you shouldn't be using them.

files.append and anything under passwd that isn't sshkeys will error (for passwd it will error the next time you update sshkeys, this is somewhat inconsistent with everything else), so we should probably actually Error not Warn for those cases.

Lastly, I think items like `unit.mask`, `files.overwrite`, `verification` etc. fields aren't actually handled by the MCO, so while they'll work for first time installation via ignition, later writes via MC probably won't work. Maybe those should be warnings? Maybe outside of the scope of this PR though so I don't feel strongly feel about this.

bgilbert:
Discussed this OOB with @yuqi-zhang.  https://github.com/coreos/fcct/pull/219#issuecomment-806177728 is correct that this is intended to support day 1.  The initial use case is improving the workflow for [boot disk mirroring support](https://docs.openshift.com/container-platform/4.7/installing/install_config/installing-customizing.html#installation-special-config-mirrored-disk_installing-customizing), which currently has to plumb FCC output directly into the pointer config.

Going through the cases discussed above:

> With that in mind and cross referencing some Openshift docs, I think `Storage.Directories` and `Storage.Links` and `Storage.Raid` is not referenced to performing any configuration Day 1, but wouldn't error if you did. We can upgrade these to Errors as well since you shouldn't be using them.

The code now errors on `Directories` and `Links`.  @yuqi-zhang and I agreed that since FCCT is more likely to be used for setting up new clusters, we have the opportunity to be more restrictive about those fields than the MCO is.

`Disks`, `Filesystems`, `Luks`, and `Raid` are all used by `boot_device` sugar, and some of those are also used directly by customers, so we shouldn't warn on them.

> files.append and anything under passwd that isn't sshkeys will error (for passwd it will error the next time you update sshkeys, this is somewhat inconsistent with everything else), so we should probably actually Error not Warn for those cases.

It looks like `Passwd.User` has the on-next-update error semantics in MCO but `Group` doesn't.  In any event, I've made these all errors, consistent with `Directories`/`Links`.

> Lastly, I think items like `unit.mask`, `files.overwrite`, `verification` etc. fields aren't actually handled by the MCO, so while they'll work for first time installation via ignition, later writes via MC probably won't work. Maybe those should be warnings? Maybe outside of the scope of this PR though so I don't feel strongly feel about this.

Looking at the MCO code, it appears that `Mask` is handled now.

The purpose of `overwrite: false` is to fail provisioning if the specified file exists in the OS image (e.g. an older config writes a file that's shipped by a newer OS).  Once we reach day 2, the MCO will keep the file in sync and `Overwrite` is no longer relevant.  So it seems reasonable not to warn on the field.

If we warn about `Verification`, people will just remove the field.  I don't think we should actively discourage users from including file digests in the config, and ignoring the field has no effect on the state of the system, so I'm inclined not to warn on that field either.

Code is updated to match the above.  Thanks for the writeup, @yuqi-zhang!

yuqi-zhang (approved):
I agree with @bgilbert's assessment. The new version LGTM. Thanks for diving in!

arithx (approved):

PR #229, opened 2021-04-05
NEWS: add news for 0.11.0

bgilbert:


NEWS
```diff
@@ -1,3 +1,35 @@
...
+
+  Breaking Changes:
+
+    - Rename project to Butane and binary to butane
+    - Change package path to github.com/coreos/butane (Go API)
+    - Remove translate.AddIdentity() in favor of translate.MergeP() (Go API)
+
+  Features:
+
+    - Add OpenShift spec 4.8.0, targeting Ignition spec 3.2.0
```
arithx, line 11: While it's redundant it might make it easier for quick skimming to still add the variant & version at the end.

arithx (approved):

miabbott (approved):

NEWS
bgilbert: I'm inclined to leave it alone, on the theory that the addition of a new spec is a global change that might be relevant info for all users.

PR #234, opened 2021-04-23
config/util: add note to all YAML output warning against editing

bgilbert:
YAML might be mistaken for a user-editable config, and it supports comments, so we might as well be clear.


arithx (approved):

bgilbert:
Overriding unrelated CI failure from GitHub Actions image regression.

PR #247, opened 2021-06-10
NEWS: add news for 0.12.1

bgilbert:


bgilbert:
Historically we've noted it only in the GitHub release.  There's an argument that the NEWS file only applies to the source release.

PR #256, opened 2021-06-24
docs: Split deprecated spec versions from stable ones

travier:
This clarifies that they should not be used and that they will be
removed in a future major release.

### DIFF
--- a/docs/specs.md
+++ b/docs/specs.md
@@ -22,8 +22,6 @@ We recommend that you always use the latest **stable** specification for your op
   - [v1.0.0](config-fcos-v1_0.md)
 - OpenShift (`openshift`)
   - [v4.8.0](config-openshift-v4_8.md)
-- RHEL CoreOS (`rhcos`) (deprecated)
-  - [v0.1.0](config-rhcos-v0_1.md)
 
 ## Experimental specification versions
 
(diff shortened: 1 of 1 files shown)


docs/specs.md
```diff
@@ -34,6 +32,13 @@ Do not use **experimental** specifications for anything beyond **development and
 - OpenShift (`openshift`)
   - [v4.9.0-experimental](config-openshift-v4_9-exp.md)
 
+## Deprecated specification versions
+
+The following configuration specifications are still understood by Butane but are considered deprecated and will be removed in a future major release:
```
bgilbert, line 37: nits:
```suggestion
The following configuration specifications are still accepted by Butane but are deprecated and will be removed in a future release:
```

bgilbert (approved):
We don't currently have concrete plans to remove the `rhcos` `0.1.0` spec, but I agree that it's a reasonable thing to do, maybe after OCP 4.7 goes EOL.  Semver says we don't need to wait for a major release because Butane is still 0.x.

bgilbert:
Filed https://github.com/coreos/butane/issues/257 to track removal of the spec.

bgilbert (approved):
:+1:

PR #261, opened 2021-07-09
Stabilize `fcos` 1.4.0 and `openshift` 4.9.0

bgilbert:


config/fcos/v1_5_exp/translate.go
```diff
@@ -57,33 +57,33 @@ const (
...
 // an error is returned.
-func (c Config) ToIgn3_3(options common.TranslateOptions) (types.Config, report.Report, error) {
-	cfg, r, err := cutil.Translate(c, "ToIgn3_3Unvalidated", options)
+func (c Config) ToIgn3_4(options common.TranslateOptions) (types.Config, report.Report, error) {
+	cfg, r, err := cutil.Translate(c, "ToIgn3_4Unvalidated", options)
 	return cfg.(types.Config), r, err
 }
 
-// ToIgn3_3Bytes translates from a v1.4 Butane config to a v3.3.0 Ignition config. It returns a report of any errors or
+// ToIgn3_4Bytes translates from a v1.5 Butane config to a v3.4.0 Ignition config. It returns a report of any errors or
```
jlebon, line 82: OOC: Is the move to targeting 3.4.0-experimental a given when creating a new experimental Butane config? I'd think it would be a more "lazy" approach where we keep targeting the same version until we have to move in case e.g. we add new Butane sugar which doesn't need new Ignition.

jlebon (approved):
LGTM!

config/fcos/v1_5_exp/translate.go
bgilbert: It's not a given.  I've generally assumed that we want to be developing against the experimental Ignition spec, and can always roll back to the stable Ignition one before stabilizing the Butane spec.  But you're right, it could make sense to go the other way.

Issue #175, opened 2020-12-07
Add sugar for kdump configuration


travier:
# What are the configuration options for kdump?

Enabling kdump on FCOS (& RHCOS) requires:
  * setting a kernel argument (for example: `crashkernel=256M`) which has a [default suggested value](https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html?highlight=kdump#boot-into-system-kernel);
  * setting options in `/etc/kdump.conf` to configure where the coredump will be stored;
  * setting options in `/etc/sysconfig/kdump` to configure additional kernel arguments passed to the "dump" kernel instance.

# What would the sugar looks like?

From https://github.com/coreos/fedora-coreos-tracker/issues/622#issuecomment-694225598:

```
kdump:
  crashkernel: 128M
  kargs: foo=bar
  config: |
    net penguin.example.com:/export/cores
```

# References

* [RHEL 8 documentation for examples](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/system_design_guide/installing-and-configuring-kdump_system-design-guide).
* See also initial discussion in https://github.com/coreos/fedora-coreos-tracker/issues/622.

bgilbert:
- Rather than `crashkernel: 128M`, maybe something like `reserved_mib: 128`?
- Is it common to want to pass additional arguments to the dump kernel?  Or can we let the user configure this themselves if they want it?

travier:
> * Rather than `crashkernel: 128M`, maybe something like `reserved_mib: 128`?

The [full option syntax](https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html?highlight=kdump#extended-crashkernel-syntax) enables setting memory size dependent values and offset but I don't know if that's something we want to support.

> * Is it common to want to pass additional arguments to the dump kernel?  Or can we let the user configure this themselves if they want it?

I think this heavily depend on where the coredump will be stored as you might need to setup networking for the crash initrd.

bgilbert:
> The full option syntax enables setting memory size dependent values and offset but I don't know if that's something we want to support.

I suspect that's a pretty unusual setting.  Users can always bypass the sugar and configure advanced settings manually.

> I think this heavily depend on where the coredump will be stored as you might need to setup networking for the crash initrd.

Would it make sense to provide target-specific sugar?

```yaml
kdump:
  reserved_mib: 128
  destinations:
    net:
      kernel_args: ip=X
      remotes:
[...]

PR #274, opened 2021-08-12
openshift/v4_10_exp: Add kdump configuration for MachineConfig

hshiina:
Enabling kdump requires much configuration such as a systemd unit, kernel arguments, configuration files. This change simplifies the configuration by introducing a syntax sugar discussed in #175 for openshift.

### DIFF
--- a/config/openshift/v4_10_exp/schema.go
+++ b/config/openshift/v4_10_exp/schema.go
@@ -36,4 +36,25 @@ type OpenShift struct {
 	Extensions      []string `yaml:"extensions"`
 	FIPS            *bool    `yaml:"fips"`
 	KernelType      *string  `yaml:"kernel_type"`
+	Kdump           Kdump    `yaml:"kdump"`
+}
+
+type Kdump struct {
+	Enabled        bool        `yaml:"enabled"`
+	ReservedMemory string      `yaml:"reserved_memory"`
+	Target         KdumpTarget `yaml:"target"`
+}
+
+type KdumpTarget struct {
+	Local Local `yaml:"local"`
+	NFS   NFS   `yaml:"nfs"`
+}
+
+type Local struct {
+	Path string `yaml:"path"`
+}
+
+type NFS struct {
+	Share string `yaml:"share"`
+	Path  string `yaml:"path"`
 }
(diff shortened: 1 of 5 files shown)


config/openshift/v4_10_exp/schema.go
```diff
@@ -36,4 +36,25 @@ type OpenShift struct {
 	Extensions      []string `yaml:"extensions"`
 	FIPS            *bool    `yaml:"fips"`
 	KernelType      *string  `yaml:"kernel_type"`
+	Kdump           Kdump    `yaml:"kdump"`
```
bgilbert, line 39: The `OpenShift` struct is only for fields [included directly in the `MachineConfigSpec`](https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfiguration.md#machineconfig-definition), outside of the Ignition config itself.  Also, since kdump configuration is useful for both Fedora CoreOS and RHEL CoreOS, this shouldn't be implemented directly in the `openshift` variant.

Instead, please move kdump support to the `fcos` 1.5.0-experimental spec.  The `openshift` 4.10.0-experimental spec will inherit it from there.  The `Kdump` field can go in the `fcos` `Config` struct.

config/openshift/v4_10_exp/translate.go
```diff
@@ -295,3 +314,75 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+
+// translateKdump converts OpenShift.Kdump field to other core fields.
+func (c *Config) translateKdump() (r report.Report) {
+	kdump := &c.OpenShift.Kdump
+	if !kdump.Enabled {
+		return r
+	}
+
+	// Enable kdump.service
+	unit := reflect.New(reflect.TypeOf(c.Systemd.Units).Elem()).Elem()
```
bgilbert, line 326: Reflection shouldn't be needed in this PR.  Instead, see `processBootDevice()` for a good way to implement this.  The idea is to build a `types.Config` data structure, then use `MergeTranslatedConfigs()` to merge it (as the parent config) with the remainder of the user-provided config (as the child config).

bgilbert (commented):
This is great, thank you!  I haven't reviewed it carefully yet, but here are two initial high-level comments.

hshiina:
Thank you for the comments. I will update the PR.

hshiina:
I'm afraid I got busy for other tasks after the last comment and have not been able to spend time on this work.

PR #276, opened 2021-08-27
main: Write output ignition files with mode 0600

cgwalters:
Because the config is commonly expected to have secret values, use mode 0600.
xref https://github.com/coreos/fedora-coreos-docs/issues/306


yasminvalim:
@cgwalters thanks for your work here!

Please, feel free to open a PR and continue to work on this issue in Ignition. I will be glad to review. 🐱

PR #278, opened 2021-09-09
config/fcos/*: add reserved partitions on aarch64/ppc64le

bgilbert:
On aarch64 and ppc64le, FCOS and RHCOS have added reserved partitions to keep partition numbers consistent between arches.  Add those partitions to the boot device mirror templates, including in stable specs.  This shouldn't introduce compatibility problems: transposefs in older OS releases will ignore the reserved partitions, any configs relying on partition numbers should be specifying them explicitly, and in any event this only affects secondary arches.

Details in https://github.com/coreos/fedora-coreos-tracker/issues/855.

### DIFF
--- a/config/fcos/v1_3/translate.go
+++ b/config/fcos/v1_3/translate.go
@@ -29,18 +29,14 @@ import (
 )
 
 const (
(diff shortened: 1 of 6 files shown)


jlebon (approved):
Looks sane to me.

Issue #243, opened 2021-06-08
Warn if partition number 0 specified with label from default OS image


bgilbert:
If the user specifies this:

```yaml
storage:
  disks:
    - device: /dev/sda
      partitions:
        - number: 0
          label: root
          size_mib: 12000
          resize: true
```

they probably meant `number: 4`.  The above will create an additional partition labeled `root`, rather than resizing the existing one.  See https://github.com/coreos/ignition/issues/1219 for details.

Add FCOS validation to check for default partlabels (`root`, `boot`, `EFI-SYSTEM`, `BIOS-BOOT`, etc.) and warn if the partition number is 0.  This is a reasonable warning even if we're not partitioning the boot disk, since udev can only create one partition symlink per partlabel for the whole system, and the user can easily choose another label.

PR #279, opened 2021-09-13
config/fcos: warn if the partition number for root isn't correct

sohankunkerkar:
Fixes https://github.com/coreos/butane/issues/243

This change warns if the partition number for the `root` label is not specified or wrongly mentioned.

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -50,6 +50,9 @@ var (
 	ErrUnknownBootDeviceLayout = errors.New("layout must be one of: aarch64, ppc64le, x86_64")
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
+	// partition
+	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+
 	// MachineConfigs
 	ErrFieldElided            = errors.New("field ignored in raw mode")
 	ErrNameRequired           = errors.New("metadata.name is required")
--- a/config/fcos/v1_3/translate.go
+++ b/config/fcos/v1_3/translate.go
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for i, disk := range ret.Storage.Disks {
+		// In the boot_device.mirror case, nothing specifies partition numbers
+		// so match existing partitions only when `wipeTable` is false
+		if disk.WipeTable == nil {
+			for j, partition := range disk.Partitions {
+				// check for reserved partlabels
+				if partition.Label != nil {
+					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", j, "label"), common.ErrWrongPartitionNumber)
+					}
+				}
+			}
+		}
+	}
 	return ret, ts, r
 }
 
--- a/config/fcos/v1_3/translate_test.go
+++ b/config/fcos/v1_3/translate_test.go
@@ -37,6 +37,7 @@ func TestTranslateBootDevice(t *testing.T) {
 		in         Config
 		out        types.Config
 		exceptions []translate.Translation
+		report     report.Report
 	}{
 		// empty config
 		{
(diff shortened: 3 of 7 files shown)


config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,13 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for _, disk := range ret.Storage.Disks {
+		for _, partition := range disk.Partitions {
+			if *partition.Label == "root" && partition.Number != 4 {
```
jlebon, line 69: I think `partition.Label` will be `nil` if no label is specified. So this needs to be e.g. 

```suggestion
			if partition.Label != nil && *partition.Label == "root" && partition.Number != 4 {
```

and similarly for the others.

config/common/errors.go
```diff
@@ -47,6 +47,9 @@ var (
 	ErrUnknownBootDeviceLayout = errors.New("layout must be one of: aarch64, ppc64le, x86_64")
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
+	// partition
+	ErrWrongPartitionNumberForRoot = errors.New("new partition using reserved label \"root\"")
```
bgilbert, line 51: This is nice and concise, but I think it'll really confuse users who didn't know they needed to set `number`.

config/fcos/v1_5_exp/translate.go
```diff
@@ -64,6 +64,13 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for _, disk := range ret.Storage.Disks {
+		for _, partition := range disk.Partitions {
+			if partition.Label != nil && *partition.Label == "root" && partition.Number != 4 {
+				r.AddOnWarn(path.New("json", "fcos", "partition"), common.ErrWrongPartitionNumberForRoot)
```
bgilbert, line 70: The path isn't arbitrary; it needs to match the actual path from the root of the output to the field.

config/fcos/v1_5_exp/translate.go
```diff
@@ -64,6 +64,13 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for _, disk := range ret.Storage.Disks {
+		for _, partition := range disk.Partitions {
+			if partition.Label != nil && *partition.Label == "root" && partition.Number != 4 {
```
bgilbert, line 69: `root` is the main one, but I think it's reasonable to check for the other reserved partitions as mentioned in https://github.com/coreos/butane/issues/243.  @jlebon, WDYT?

bgilbert (commented):
Will need tests before landing.

config/fcos/v1_5_exp/translate.go
sohankunkerkar: `path.New("json", "storage", "disks", 0, "device", "partitions", 0, "number")` ?

bgilbert: More or less, yeah.  `device` isn't in the path, and instead of zeroes, we need to use the indexes of the array entries.

config/fcos/v1_5_exp/translate.go
jlebon: Yup, that makes sense to me!

config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for i, disk := range ret.Storage.Disks {
+		// In the boot_device.mirror case, nothing specifies partition numbers
+		// so match existing paritions only when `wipeTable` is false
```
bgilbert, line 69: `partitions`

config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
...
+		if disk.WipeTable != nil && !*disk.WipeTable {
```
bgilbert, line 70: This ignores the case where `wipeTable` is not specified.

config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
...
+			for j, partition := range disk.Partitions {
+				// check for default partlabels
```
bgilbert, line 72: `reserved` is probably a bit clearer than `default`.

config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
...
+				if partition.Label != nil {
+					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", j, *partition.Label), common.ErrWrongPartitionNumber)
```
bgilbert, line 75: Paths are about field names/indexes, not values, so the last argument should just be `"label"`.

config/fcos/v1_3/translate_test.go
```diff
@@ -49,6 +50,113 @@ func TestTranslateBootDevice(t *testing.T) {
...
+										Label:   util.StrToPtr("root"),
+										SizeMiB: util.IntToPtr(12000),
+										Resize:  util.BoolToPtr(true),
+									},
+									{
+										Label:   util.StrToPtr("var-home"),
+										SizeMiB: util.IntToPtr(10240),
+									},
+								},
+								WipeTable: util.BoolToPtr(false),
```
bgilbert, line 74: It's more likely that this will be unspecified, so let's omit it.

config/fcos/v1_3/translate_test.go
```diff
@@ -49,6 +50,113 @@ func TestTranslateBootDevice(t *testing.T) {
...
+				{path.New("yaml", "storage", "filesystems", 0, "label"), path.New("json", "storage", "filesystems", 0, "label")},
+				{path.New("yaml", "storage", "filesystems", 0, "wipe_filesystem"), path.New("json", "storage", "filesystems", 0, "wipeFilesystem")},
+				{path.New("yaml", "storage", "filesystems", 0), path.New("json", "storage", "filesystems", 0)},
+				{path.New("yaml", "storage", "filesystems"), path.New("json", "storage", "filesystems")},
+				{path.New("yaml", "storage"), path.New("json", "storage")},
+			},
+			report.Report{
+				Entries: []report.Entry{
+					{
+						Kind:    report.Kind(1),
```
bgilbert, line 144: `report.Warn`

config/fcos/v1_3/translate_test.go
```diff
@@ -49,6 +50,113 @@ func TestTranslateBootDevice(t *testing.T) {
...
+						Message: "incorrect partition number; a new partition will be created using reserved label",
```
bgilbert, line 145: `errors.ErrWrongPartitionNumber.Error()` will avoid test breakage if the error message is changed.

config/fcos/v1_3/translate_test.go
```diff
@@ -49,6 +50,113 @@ func TestTranslateBootDevice(t *testing.T) {
...
+						Context: path.ContextPath{
+							Path: []interface{}{
```
bgilbert, line 147: `path.New(...)` would be clearer, and we don't need to test the internals of `path` here.

config/fcos/v1_3/translate_test.go
```diff
@@ -49,6 +50,113 @@ func TestTranslateBootDevice(t *testing.T) {
...
+								"storage",
+								"disks",
+								0,
+								"partitions",
+								0,
+								"root",
```
bgilbert, line 153: `"label"`

config/fcos/v1_3/translate.go
```diff
@@ -64,6 +64,20 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 		return types.Config{}, translate.TranslationSet{}, r
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
+	for i, disk := range ret.Storage.Disks {
+		// In the boot_device.mirror case, nothing specifies partition numbers
+		// so match existing partitions only when `wipeTable` is false
+		if (disk.WipeTable == nil) || (disk.WipeTable != nil && !*disk.WipeTable) {
```
bgilbert, line 70: The second `WipeTable` check is redundant.

bgilbert (approved):

config/fcos/v1_3/translate.go
bgilbert: ...sorry, I phrased that poorly; all three of them are `WipeTable` checks.  I meant the second nil check.  So this should be:

```go
if disk.WipeTable == nil || !*disk.WipeTable {
```

which is actually just

```go
if !util.IsTrue(disk.WipeTable) {
```

bgilbert: Fix in https://github.com/coreos/butane/pull/288.

PR #281, opened 2021-10-14
rhcos: add deprecation warning

bgilbert:
#256 marked the `rhcos` variant deprecated, but we're still seeing folks try to use it in new applications.  Add a warning when the variant is used.

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -27,6 +27,9 @@ var (
 	ErrInvalidSourceConfig    = errors.New("source config is invalid")
 	ErrInvalidGeneratedConfig = errors.New("config generated was invalid")
 
+	// deprecated variant/version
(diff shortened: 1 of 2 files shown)


sohankunkerkar (approved):

travier (approved):

config/common/errors.go
```diff
@@ -27,6 +27,9 @@ var (
 	ErrInvalidSourceConfig    = errors.New("source config is invalid")
 	ErrInvalidGeneratedConfig = errors.New("config generated was invalid")
 
+	// deprecated variant/version
+	ErrRhcosVariantDeprecated = errors.New("this variant is deprecated and will be removed in a future release; use openshift instead")
```
jlebon, line 31: Bikeshed, maybe:

```suggestion
	ErrRhcosVariantDeprecated = errors.New("this variant is deprecated and will be removed in a future release; use openshift variant instead")
```

just so it's clear we're not talking about some other OpenShift tool.

PR #291, opened 2021-11-17
docs/config-*: ignition section is optional

bgilbert:
It's mandatory for Ignition configs, but optional for Butane configs.

Reported by @Nemric.


jmarrero (approved):
lgtm

PR #294, opened 2021-12-02
docs: Don't use '--tty-' for butane alias

travier:
Using '--tty' conflicts with stdin redirection:
https://docs.podman.io/en/latest/markdown/podman-run.1.html#tty-t

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -32,7 +32,7 @@ podman run --rm -v /path/to/your_config.bu:/config.bu:z quay.io/coreos/butane:re
 You may also add the following alias in your shell configuration:
 
 ```
-alias butane='podman run --rm --tty --interactive \
+alias butane='podman run --rm --interactive       \
               --security-opt label=disable        \
               --volume ${PWD}:/pwd --workdir /pwd \
               quay.io/coreos/butane:release'


bgilbert (approved):

PhrozenByte:
Thinking about this, why is Butane recommending a Bash alias in the first place? Editing `~/.bash_aliases` or creating a new `~/.local/bin/butane` makes no big difference, and a shell script

```shell
#!/bin/sh
podman run --rm --interactive --security-opt label=disable --volume ${PWD}:/pwd --workdir /pwd quay.io/coreos/butane:release "$@"
```

is more portable considering the use in scripts. Or shall we document both?

travier:
> Thinking about this, why is Butane recommending a Bash alias in the first place?

We are *suggesting* a shell alias. You can use the option you prefer.

> Editing `~/.bash_aliases` or creating a new `~/.local/bin/butane` makes no big difference, and a shell script is more portable considering the use in scripts. Or shall we document both?

We could also document that. Is the `~/.local/bin/` path commonly set in `$PATH` in most Linux distributions?

PhrozenByte:
> Is the `~/.local/bin/` path commonly set in `$PATH` in most Linux distributions?

`~/.local/bin/` is defined in the Freedesktop.org file system hierarchy standard, see https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#~/.local/bin/. I can confirm its usage for Debian, Ubuntu (I *presume* this also implies most of its derivates) and FCOS (I *presume* this also implies Fedora, RHEL, CentOS). It kinda depends on what "most" is, e.g. Arch won't set it by default, but since the Freedesktop.org standards are pretty well adapted, I'd say that "most" distributions include it, yes. But there's no harm in adding a sentence like:

> Make sure `~/.local/bin` is in your `$PATH`, or choose another path like `/usr/local/bin`.

travier:
Great! Can you make a followup PR to include that in the docs? Thanks!

Issue #87, opened 2020-03-23
Can't run fcct 04.0 standalone binary in alpine container


jensrutschmann:
Hi,

I cannot run the latest release 0.4.0 binary inside an alpine container:
```
jens:~> docker run --rm -it alpine:3.8 sh
/ # wget -O /usr/local/bin/fcct https://github.com/coreos/fcct/releases/download/v0.4.0/fcct-x86_64-unknown-linux-gnu
Connecting to github.com (140.82.118.4:443)
Connecting to github-production-release-asset-2e65be.s3.amazonaws.com (52.216.170.59:443)
fcct                 100% |*******************************************************************************************************************************************************************************************| 10203k  0:00:00 ETA
/ # chmod a+x /usr/local/bin/fcct 
/ # /usr/local/bin/fcct 
sh: /usr/local/bin/fcct: not found
```

After installing `file` (using `apk --update-cache --upgrade add file`) it turns out the 0.4.0 release is a dynamically linked binary, not a static binary.
```
/usr/local/bin # file fcct
fcct: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/l, BuildID[sha1]=d9bb68fd48bf269bb0271cfc71545251d8fe1343, for GNU/Linux 3.2.0, stripped
```
Yet there aren't any linker issues:
```
/usr/local/bin # ldd fcct 
	/lib64/ld-linux-x86-64.so.2 (0x7f40e5620000)
	libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f40e5620000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f40e5620000)
```

Shouldn't the 0.4.0 release also be a statically linked binary?  The 0.2.0 release is, and I can run that one just fine.

tzawlxih:
6th release of fcct, still not working. 

ct-*-x86_64-unknown-linux-gnu was working well

Okeanos:
So I just stumbled across this as well – for some reason the releases are all dynamically linked instead of statically linked:

```sh
$ ldd butane
./butane: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./butane)
	linux-vdso.so.1 (0x00007ffd19c9d000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f575bd9b000)
[...]

PR #298, opened 2022-01-05
docs: add information about the build system

Okeanos:
As visible in #87: it's not very straight forward to understand where and how the final binaries offered here are built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,5 +3,7 @@
 Butane (formerly the Fedora CoreOS Config Transpiler, FCCT) translates human readable Butane Configs
 into machine readable [Ignition](https://github.com/coreos/ignition) Configs. See the [getting
 started](docs/getting-started.md) guide for how to use Butane and the [configuration
-specifications](docs/specs.md) for everything Butane configs support. For information on developing Butane
-or using it as a library, see the [development docs](docs/development.md).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,3 +5,10 @@ into machine readable [Ignition](https://github.com/coreos/ignition) Configs. Se
 started](docs/getting-started.md) guide for how to use Butane and the [configuration
 specifications](docs/specs.md) for everything Butane configs support. For information on developing Butane
 or using it as a library, see the [development docs](docs/development.md).
+
+Note that the binaries released here are not built using the `build` script from this repository but using a
```
bgilbert, line 9: Maybe "released in this repository"?  The container binary is built in Quay using `build_for_container`.

bgilbert (commented):
This is helpful, thanks!  Let's move it to the development docs though.  We already link to those from the README, and the build process is an implementation detail that won't affect most users.

If you're up for it, [Ignition](https://github.com/coreos/ignition) could use a similar note in its dev docs, since the shipped `ignition-validate` binary is likewise built in Fedora.

README.md
Okeanos: I updated the pull request accordingly :)

bgilbert (approved):
Thanks!

PR #299, opened 2022-01-07
openshift/*: reject `storage.files.contents.source` with non-`data` URL

bgilbert:
The MCO fails on a parse error when attempting to reconcile a file sourced from a URL scheme other than `data`.  Reject such URLs, including in stable specs.

Drop `http_headers` from the spec docs, since it doesn't make sense for data URLs.  Ignition config validation already rejects the field in this case.

cc @yuqi-zhang to confirm that this is the right thing to do.


sohankunkerkar (approved):

yuqi-zhang:
Sorry for the delay. This is the right path I believe. Like @bgilbert says, we only explicitly parse for `data`, and as far as I'm aware have no plans to support any other form of content sourcing in the near future.

bgilbert:
Great, thanks @yuqi-zhang!

PR #303, opened 2022-01-13
Stabilize openshift 4.10.0

sohankunkerkar:


bgilbert:
Let's copy `v4_10/` to `v4_11_exp/` in a dedicated commit and then make the necessary changes afterward.  The diff for the fourth commit isn't really readable right now.

docs/upgrading-openshift.md
```diff
@@ -13,6 +13,31 @@ Occasionally, changes are made to OpenShift Butane configs (those that specify `
 1. TOC
 {:toc}
 
+## From Version 4.9.0 to 4.10.0
+
+There are no breaking changes between versions 4.9.0 and 4.10.0 of the `openshift` configuration specification. Any valid 4.9.0 configuration can be updated to a 4.10.0 configuration by changing the version string in the config. 
+
+### File compression
+
+File compression is re-introduced in 4.10.0, and it's allowed to set `compression: gzip` in a config. Butane will automatically compress the data included with `inline` or `local`, however, it's required to set the `compression` field to explicilty compress the file content from remote URLs.
```
bgilbert, line 22: Hmm, we never documented the disablement outside Butane release notes.  Maybe something like:

> Resource compression, which was disabled in all `openshift` specs in Butane 0.12.1, is re-introduced in this spec version.  The `compression` field can be set to `gzip` to decompress gzip-compressed resources.  In addition, Butane may automatically compress resources specified with `inline` or `local`.

config/openshift/v4_11_exp/translate.go
```diff
@@ -0,0 +1,293 @@
...
+
+package v4_11_exp
+
+import (
+	"net/url"
+	"reflect"
+	"strings"
+
+	"github.com/coreos/butane/config/common"
+	"github.com/coreos/butane/config/openshift/v4_10/result"
```
bgilbert, line 23: Looks like a couple of these snuck into the diff.

bgilbert (approved):

Issue #300, opened 2022-01-09
Add the ability to define packages to be layered.


jmarrero:
As part of the layering initiative we would like to be able to declare packages as part of the schema. In the background a service that calls rpm-ostree to the installation of the packages should be created.

lucab:
This is coming from https://github.com/coreos/fedora-coreos-tracker/issues/1054.

https://github.com/coreos/butane/issues/81 has some prior discussion and examples on it.

bgilbert:
If we want to eventually support this in both the `fcos` and `openshift` specs, note that we'll need to get https://github.com/openshift/machine-config-operator/pull/2675 (or something like it) unstuck.  `openshift` specs derive their Ignition spec version from the `fcos` version they inherit from, and `openshift` stable specs are currently pinned to `fcos` 1.3.0 because the MCO currently maxes out at Ignition spec 3.2.0.

jmarrero:
@jlebon during your demo today your example was dropping a ltrace.yaml file. I assume we don't want a file per package right? @cgwalters I think suggested a SHA of the file. We could also use a UUID.yaml as filename. If I just output to `/etc/rpm-ostree/origin.d/SHAorUUID.yaml` that is what would be the "server side" expected file to output on #304 ?
```
packages:
   - strace
   - usbguard
```

Also, unless I misunderstood you suggested that we did not need the systemd service with the treefile approach on the client side. But even if we are adding a treefile using `ignition --live-apply` we still need to call `rpm-ostree` after boot to process the treefile so we still need a service no?

jlebon:
> @jlebon during your demo today your example was dropping a ltrace.yaml file. I assume we don't want a file per package right? @cgwalters I think suggested a SHA of the file. We could also use a UUID.yaml as filename.

No strong opinion on filename. Indeed we should just combine them in a single file instead of separate ones per package. I'd go for something like `extensions.yaml` to keep it simple and have some traceability. Might also be worth including a YAML comment at the top of the file to say it was generated by Butane.

> If I just output to `/etc/rpm-ostree/origin.d/SHAorUUID.yaml` that is what would be the "server side" expected file to output on #304 ?
> 
> ```
> packages:
>    - strace
>    - usbguard
> ```

LGTM

> Also, unless I misunderstood you suggested that we did not need the systemd service with the treefile approach on the client side. But even if we are adding a treefile using `ignition --live-apply` we still need to call `rpm-ostree` after boot to process the treefile so we still need a service no?

[...]

PR #304, opened 2022-01-17
config/fcos: Add extensions param and generate treefile with packages

jmarrero:
This PR adds the sugar to take a list of extensions (packages) on the butane config and turns them into a [treefile](https://coreos.github.io/rpm-ostree/treefile/) which rpm-ostree will consume as part of the derivation/layering [enhancement](https://github.com/coreos/enhancements/pull/7). This specific PR is raised based on the discussions on: https://github.com/coreos/fedora-coreos-tracker/issues/1054


jmarrero:
@bgilbert is the way I marshaling the YAML correct. It's ok to add that dependency of `yaml.v2@v2.2.2` ?

jmarrero:
Still seeing: ```Missing paths in TranslationSet:
$.storage.files.2.path
$.storage.files.2.contents.source
$.storage.files.2.contents
$.storage.files.2.mode
$.storage.files.2``` after merges trying to figure it out.

bgilbert:
> is the way I am marshaling the YAML correct? It's OK to add that dependency of yaml.v2@v2.2.2 or am I missing some obvious util?

We already have a dependency on `yaml.v3`; can you use that?

> Missing paths in TranslationSet:

What does `butane -D` say?

jmarrero:
> > is the way I am marshaling the YAML correct? It's OK to add that dependency of yaml.v2@v2.2.2 or am I missing some obvious util?
> 
> We already have a dependency on `yaml.v3`; can you use that?
> 
Yeah yaml.v3 works, I did not think of searching the other versions in the imports we are already using. Thanks.

> > Missing paths in TranslationSet:
> 
> What does `butane -D` say?


That output comes from butane -D.
The full output looks like:
```
TranslationSet: yaml → json
$.ignition → $.ignition
$.version → $.ignition.version
$.storage → $.storage
$.storage.files → $.storage.files
$.storage.files.0 → $.storage.files.0
$.storage.files.0.contents → $.storage.files.0.contents
$.inline → $.storage.files.0.contents.source
$.storage.files.0.mode → $.storage.files.0.mode
$.storage.files.0.path → $.storage.files.0.path
$.storage.files.1 → $.storage.files.1
$.storage.files.1.contents → $.storage.files.1.contents
$.storage.files.1.contents.inline → $.storage.files.1.contents.compression
$.storage.files.1.contents.inline → $.storage.files.1.contents.source
$.storage.files.1.mode → $.storage.files.1.mode
$.storage.files.1.path → $.storage.files.1.path
$.systemd → $.systemd
$.systemd.units → $.systemd.units
$.systemd.units.0 → $.systemd.units.0
$.systemd.units.0.dropins → $.systemd.units.0.dropins
$.systemd.units.0.dropins.0 → $.systemd.units.0.dropins.0
$.systemd.units.0.dropins.0.contents → $.systemd.units.0.dropins.0.contents
$.systemd.units.0.dropins.0.name → $.systemd.units.0.dropins.0.name
$.systemd.units.0.name → $.systemd.units.0.name
$.systemd.units.1 → $.systemd.units.1
$.systemd.units.1.contents → $.systemd.units.1.contents
$.systemd.units.1.enabled → $.systemd.units.1.enabled
$.systemd.units.1.name → $.systemd.units.1.name

Missing paths in TranslationSet:
$.storage.files.2.path
$.storage.files.2.contents.source
$.storage.files.2.contents
$.storage.files.2.mode
$.storage.files.2
```

Which if I understand is because the translation set from the procressPackages() sets the file at index 0 but after appending it is of index 2. If that is correct do I need to modify that manually or is there other method I am missing that is better than merge to deal with this?

config/fcos/v1_5_exp/translate.go
```diff
@@ -78,6 +80,14 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
 			}
 		}
 	}
+
+	retp, rp, tsp := c.processPackages(options)
+
+	ret.Storage.Files = append(ret.Storage.Files, retp.Storage.Files...)
```
bgilbert, line 86: You can't just glue some arrays together here; you need to run a proper config merge with `MergeTranslatedConfigs`.  That'll take care of mapping the translation set indices for you.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,51 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
 	tm.AddTranslation(path.New("yaml"), path.New("json"))
 	return
 }
+
+func (c Config) processPackages(options common.TranslateOptions) (types.Config, report.Report, translate.TranslationSet) {
+	yamlPath := path.New("yaml", "inline")
```
bgilbert, line 307: In struct translation functions, the path is relative to the root of the struct.  In desugaring functions, though, it's relative to the root of the config.  So this isn't correct; you need a full path here.

bgilbert (commented):
You'll want `MergeTranslatedConfigs`; see below.

config/fcos/v1_5_exp/translate.go
jmarrero: I am not entirely sure I get which config. My latest commit I assumes is the config the user provides not the types.Config right?. So `yamlPath := path.New("yaml", "extensions")` ?

bgilbert: Right, the root of the input (YAML, Butane) config.

jmarrero:
Working on a test now. Thanks for all the help @bgilbert

config/fcos/v1_5_exp/schema.go
```diff
@@ -21,6 +21,7 @@ import (
 type Config struct {
 	base.Config `yaml:",inline"`
 	BootDevice  BootDevice `yaml:"boot_device"`
+	Extentions  Extensions `yaml:"extensions"`
```
bgilbert, line 24: `Extensions`

config/fcos/v1_5_exp/translate.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/coreos/butane/config/common"
 	cutil "github.com/coreos/butane/config/util"
 	"github.com/coreos/butane/translate"
+	"gopkg.in/yaml.v3"
```
bgilbert, line 24: This block is for Butane packages.  External dependencies should be in the next one.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
 	tm.AddTranslation(path.New("yaml"), path.New("json"))
 	return
 }
+
+func (c Config) processPackages(options common.TranslateOptions) (types.Config, report.Report, translate.TranslationSet) {
```
bgilbert, line 306: Whoops, I had the return order wrong.  Looks like the report should come last.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+
+	type TreeFile struct {
+		Packages []string `yaml:"packages"`
+	}
+
+	p := TreeFile{
+		Packages: c.Extentions,
+	}
+
+	marshalledYAML, err := yaml.Marshal(&p)
```
bgilbert, line 325: Could be simplified to an anonymous struct:

```go
	treeFileContents, err := yaml.Marshal(&struct {
		Packages []string `yaml:"packages"`
	}{
		Packages: c.Extensions,
	})
```

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+	if err != nil {
+		r.AddOnError(yamlPath, err)
+		return ret, r, ts
+	}
+
+	treeFilePath := "/etc/rpm-ostree-layer.yaml"
+
+	file := types.File{
+		Node: types.Node{
+			Path: treeFilePath,
```
bgilbert, line 336: Cleanup: the path literal could be inlined here.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+	}
+	if gzipped {
+		file.Contents.Compression = util.StrToPtr("gzip")
+	}
+	file.Contents.Source = util.StrToPtr(src)
+
+	file.Mode = util.IntToPtr(0644)
+
+	ret.Storage.Files = append(ret.Storage.Files, file)
+	ts.AddFromCommonSource(yamlPath, path.New("json", "storage", "files", 0), file)
```
bgilbert, line 353: This fails to add paths for `storage` and `storage.files`.
```suggestion
	ts.AddFromCommonSource(yamlPath, path.New("json", "storage"), ret.Storage)
```

config/fcos/v1_5_exp/translate.go
```diff
@@ -78,6 +79,15 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
...
 		}
 	}
+
+	retp, rp, tsp := c.processPackages(options)
+
+	retConfig, retTranslations := baseutil.MergeTranslatedConfigs(retp, tsp, ret, ts)
+	ret = retConfig.(types.Config)
+	ts = retTranslations
+
+	r.Merge(rp)
```
bgilbert, line 89: Cleanup: let's keep all of these lines in one block to show they're related.

config/fcos/v1_5_exp/translate.go
```diff
@@ -78,6 +79,15 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
 			}
 		}
 	}
+
+	retp, rp, tsp := c.processPackages(options)
+
+	retConfig, retTranslations := baseutil.MergeTranslatedConfigs(retp, tsp, ret, ts)
```
bgilbert, line 85: I'd be tempted to spell this

```go
var retConfig types.Config
retConfig, ts = ...
```

but I can see it going either way.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+
+	treeFilePath := "/etc/rpm-ostree-layer.yaml"
+
+	file := types.File{
+		Node: types.Node{
+			Path: treeFilePath,
+		},
+	}
+
+	src, gzipped, err := baseutil.MakeDataURL([]byte(marshalledYAML), file.Contents.Compression, !options.NoResourceAutoCompression)
```
bgilbert, line 340: The second argument is always nil, so for clarity I think it's better to pass nil there.

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +302,55 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+	if err != nil {
+		r.AddOnError(yamlPath, err)
+		return ret, r, ts
+	}
+	if gzipped {
+		file.Contents.Compression = util.StrToPtr("gzip")
+	}
+	file.Contents.Source = util.StrToPtr(src)
+
+	file.Mode = util.IntToPtr(0644)
```
bgilbert, line 350: Cleanup: I'd probably move the `file` construction down here, and set the fixed fields as part of construction.

config/fcos/v1_5_exp/translate.go
jmarrero: If we know that it's always going be nil and gzipped is gonna always be false, any reason to not just ignore `gzipped` and turn it into `_`?

bgilbert: `gzipped` won't always be false.  `MakeDataURL` makes a decision about whether to compress.

bgilbert: The actual rule is: skip compression if the third argument is `false` or if the second argument is `&"gzip"`.

jmarrero:
Moving back to WIP. This depends on the outcome of: 
https://github.com/coreos/rpm-ostree/pull/3340

config/fcos/v1_5_exp/schema.go
```diff
@@ -38,3 +39,5 @@ type BootDeviceLuks struct {
 type BootDeviceMirror struct {
 	Devices []string `yaml:"devices"`
 }
+
+type Extensions []string
```
bgilbert, line 43: Are we sure we won't have additional options related to extensions in the future?  This schema doesn't allow for any.

The usual way we do this is:
```yaml
extensions:
  - name: foo
  - name: bar
```

config/fcos/v1_5_exp/translate.go
```diff
@@ -292,3 +300,50 @@ func translateBootDeviceLuks(from BootDeviceLuks, options common.TranslateOption
...
+	treeFileContents, err := yaml.Marshal(&struct {
+		Packages []string `yaml:"packages"`
+	}{
+		Packages: c.Extensions,
+	})
+	if err != nil {
+		r.AddOnError(yamlPath, err)
+		return ret, ts, r
+	}
+	fullYamlContents := append([]byte("###File Generated by Butane\n"), treeFileContents...)
```
bgilbert, line 322: Bikeshed:
```suggestion
	fullYamlContents := append([]byte("# Generated by Butane\n\n"), treeFileContents...)
```

bgilbert (commented):
LGTM generally, pending tests.

config/fcos/v1_5_exp/schema.go
jmarrero: We are basically adding something like the RHCOS extensions to butane: https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfiguration.md#rhcos-extensions .  But I am not sure how this will evolve. This is still being discussed but for the first pass AFAIK we just care about a list of packages.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1417,3 +1417,59 @@ func TestTranslateBootDevice(t *testing.T) {
...
+
+// TestTranslateExtensions tests translating the Butane config extensions section.
+func TestTranslateExtensions(t *testing.T) {
+	tests := []struct {
+		in         Config
+		out        types.Config
+		exceptions []translate.Translation
+		report     report.Report
+	}{
+		// empty config
```
bgilbert, line 1429: This is not an empty config.  :slightly_smiling_face:

docs/config-fcos-v1_5-exp.md
```diff
@@ -203,6 +203,7 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_threshold_** (int): sets the minimum number of pieces required to decrypt the device. Default is 1.
   * **_mirror_** (object): describes mirroring of the boot disk for fault tolerance.
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
+* **extensions** (list of strings): A list of packages to layer on top of the OS.
```
bgilbert, line 206: lowercase A.

Also, we should either a) add this to the OpenShift experimental spec docs, or (b) have the OpenShift spec validator explicitly reject the field.

config/fcos/v1_5_exp/schema.go
bgilbert: Once this spec version has been stabilized, we lose the ability to add metadata in this way without making a new top-level field.  We could merge this as-is and revisit before stabilization, but if we do, we should make sure to leave clear signposts in the issue tracker so we don't forget.  Or we could go for the extensible approach now.  It might be worth raising with the team.

config/fcos/v1_5_exp/translate_test.go
jmarrero: Wooops. Left behind from copy/paste from the original test. 😀 addressing asap

docs/config-fcos-v1_5-exp.md
jmarrero: Is there a best practice on which to select? Specially if this could be changing?

bgilbert: It's safer to disable the functionality in `openshift` and possibly enable it later, rather than the other way around.

Now that I think about it, `openshift` configs already have a way to specify OS extensions (`openshift.extensions`).  The last time we added that sort of duplication (for kernel arguments) we had `openshift` disable the fields inherited from the `fcos` variant, and continued to support only the `openshift`-specific fields there.

config/fcos/v1_5_exp/schema.go
jmarrero: @cgwalters @jlebon any preference on how to handle this?

docs/config-fcos-v1_5-exp.md
jmarrero: This might tie into this thread: https://github.com/coreos/butane/pull/304#discussion_r790336682 we want this to be fcos + openshift thing. So it might require another chat with the MCO team. I am more leaning into creating a Issue that says to revisit this once it's time to stabilize the API. I think for now we need to allow both on openshift.

bgilbert: Well, it also ties into https://github.com/coreos/butane/issues/300#issuecomment-1015041837.  Even if `openshift` inherits it now, that'll change when 4.11.0 is stabilized unless the MCO starts supporting Ignition spec 3.3.0.

So I'm okay leaving it enabled for now and filing a bug.  I've created stabilization milestones for the next `fcos` and `openshift` specs in the issue tracker, and we can attach to those.  Could you handle filing the issue?

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1417,3 +1417,59 @@ func TestTranslateBootDevice(t *testing.T) {
...
+				{path.New("yaml", "extensions"), path.New("json", "storage", "files", 0, "path")},
+				{path.New("yaml", "extensions"), path.New("json", "storage", "files", 0, "mode")},
+				{path.New("yaml", "extensions"), path.New("json", "storage", "files", 0, "contents")},
+				{path.New("yaml", "extensions"), path.New("json", "storage", "files", 0, "contents", "source")},
+			},
+			report.Report{},
+		},
+	}
+
+	for i, test := range tests {
```
bgilbert, line 1468: e83e1c99d8e45b7f8e0add5dffaef9751d34370f introduced a style change for these loops.  Could you apply the update when you next respin?

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1417,3 +1417,59 @@ func TestTranslateBootDevice(t *testing.T) {
...
+		actual, translations, r := test.in.ToIgn3_4Unvalidated(common.TranslateOptions{})
+		assert.Equal(t, test.out, actual, "#%d: translation mismatch", i)
+		assert.Equal(t, test.report, r, "#%d: report mimatch", i)
```
bgilbert, line 1471: Copypasted typo: `mismatch`.  I'll handle fixing the original.

docs/config-fcos-v1_5-exp.md
bgilbert: In that case, please do update the `openshift` spec docs as well.

config/fcos/v1_5_exp/translate_test.go
bgilbert: #313

config/fcos/v1_5_exp/schema.go
jlebon: Agree with the concern, but not sure what to recommend since it's hard to predict what we'll need to generalize (container-based extensions? more treefile support? repo-pinned extensions?).

So... I'd say let's roll with this for now and revisit before stabilization. In the worst case, we should be able to change the spec in future versions, right?

docs/config-fcos-v1_5-exp.md
jmarrero: Opened: https://github.com/coreos/butane/issues/314

config/fcos/v1_5_exp/translate_test.go
jmarrero: done

config/fcos/v1_5_exp/schema.go
bgilbert: Discussed this more with @jlebon OOB.  If we wanted to add extensibility in future versions, and didn't want to have a major spec bump (i.e. one that requires config changes when bumping the version number), we'd need to deprecate the `extensions` field and add e.g. `extensions2`.  We've done that before but it's a bit awkward.

I think the maximum possible extensibility is:

```yaml
extensions:
  packages:
    - name: foo
    - name: bar
```

which lets us add system-wide extension options, per-package ones, and new extension types:

```yaml
extensions:
  global-option: z
  packages:
    - name: foo
      package-option: z
    - name: bar
  containers:
    - name: baz
```

Global options don't play well with config merging, since a child config that adds a package would be affected by global parameters it isn't aware of.  New extension types could be implemented as a package field:

```yaml
extensions:
  - name: foo
    package-option: z
  - name: bar
  - name: baz
    type: container
```

which is consistent with the design elsewhere in the schema.

I think that's a reasonable middle ground.  Let's not block the current PR on this, but as a followup I think we should switch from an array of strings to an array of structs.

bgilbert (approved):

config/fcos/v1_5_exp/schema.go
cgwalters: This is very related to what happened to `packages` in 
https://github.com/coreos/rpm-ostree/blob/main/docs/treefile.md

We ended up adding `repo-packages` which allows to pin to packages from a specific repo.

--

The idea of `type: container` is a really interesting idea.  Definitely touches on https://www.freedesktop.org/software/systemd/man/systemd-sysext.html etc.

cgwalters:
One concern @jmarrero mentioned in a chat about this is that we're implicity supporting whatever CLI arguments can be passed to `rpm-ostree install`.  Today that includes e.g. bodhi update URLs as well as direct e.g. `https://example.com/foo.rpm`.

I think that's probably OK.

This also relates though to a discussion happening in libdnf 
https://github.com/rpm-software-management/libdnf/issues/1375

cgwalters:
Did the question come up of potentially having butane generate default configuration files or kernel arguments when a specific extension is requested?  `kdump` is a great example of this.  https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/

There may be others where we'd want to drop some config file in `/etc`.

IOW, should butane ever try to recognize specific extensions, or should we just pass through to `rpm-ostree install`.

config/fcos/v1_5_exp/schema.go
cgwalters: In rpm-ostree's treefiles today to be nice we allow multiple packages in single YAML list entry.  See for example https://github.com/coreos/fedora-coreos-config/blob/aa50439847f34282b41bf2eb60d478a45d2a27d7/manifests/system-configuration.yaml#L7

Might make sense to do that too to avoid the overhead in the simple cases?

bgilbert: We generally don't do that in _Ignition_ configs because it wouldn't work with the config merging logic, which (with one unfortunate exception) doesn't know the semantics of what it's merging.  That argument doesn't directly apply here, since this is sugar, but it still doesn't feel very Butane-like.  We've favored explicitness over ergonomics when they're in conflict.

bgilbert:
> One concern @jmarrero mentioned in a chat about this is that we're implicity supporting whatever CLI arguments can be passed to rpm-ostree install. Today that includes e.g. bodhi update URLs as well as direct e.g. https://example.com/foo.rpm.

Can we filter those out e.g. by rejecting slashes?  Explicit > implicit, and we can always loosen validation rules later.

> IOW, should butane ever try to recognize specific extensions, or should we just pass through to rpm-ostree install.

If certain extensions need additional configuration, we should add explicit sugar for those.  Otherwise, validation has no way to know that a particular extension will invoke additional logic in a future Butane release, so a newer Butane config could produce a subtly broken Ignition config with older Butane.

config/fcos/v1_5_exp/schema.go
bgilbert: Followup in #316.

cgwalters:
> Can we filter those out e.g. by rejecting slashes? Explicit > implicit, and we can always loosen validation rules later.

Actually I realized I was wrong, since the change to generate a treefile instead of forking `rpm-ostree install`, I think only package names/NEVRAs will be supported.  (The direct `https://example.com/foo.rpm` stuff lives in the client binary today)

jmarrero:
https://github.com/coreos/butane/pull/304 & https://github.com/coreos/butane/pull/316 solve this issue.

PR #311, opened 2022-01-23
Cleanup

Okeanos:
While working on #289 I stumbled across a number of nitpicks and things I would've done differently based on the supported Go version (1.15 in go.mod).

Grouped by individual commits in this pull request I did the following things:

- fix some (obvious) typos
- use `t.TempDir` in tests: a Go managed temporary directory automatically cleaned up at the end of tests, available since Go 1.15
- Prevent shadowing of imports
- Remove unnecessary typing of particular variable declarations
- Blank some unused method parameters
- Use Go subtests (available since Go 1.7) instead of simple loops. This allows better debugging in case of errors because the faulty iteration will now tell us which input in particular failed while running through all of them at all times.

### DIFF
--- a/base/util/merge_test.go
+++ b/base/util/merge_test.go
@@ -15,6 +15,7 @@
 package util
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/coreos/butane/translate"
@@ -141,9 +142,11 @@ func TestMergeTranslatedConfigs(t *testing.T) {
 		},
 	}
 	for i, test := range tests {
-		c, ts := MergeTranslatedConfigs(test.parent, test.parentTranslations, test.child, test.childTranslations)
-		assert.Equal(t, test.merged, c, "#%d: bad config", i)
-		assert.Equal(t, test.mergedTranslations, ts, "#%d: bad translations", i)
+		t.Run(fmt.Sprintf("merge %d", i), func(t *testing.T) {
+			c, ts := MergeTranslatedConfigs(test.parent, test.parentTranslations, test.child, test.childTranslations)
+			assert.Equal(t, test.merged, c, "bad config")
+			assert.Equal(t, test.mergedTranslations, ts, "bad translations")
+		})
 	}
 }
 
--- a/base/util/file_test.go
+++ b/base/util/file_test.go
@@ -101,12 +101,15 @@ var (
 func TestCheckForDecimalMode(t *testing.T) {
 	// test decimal to octal conversion
 	for i := -1; i < 10001; i++ {
-		iStr := fmt.Sprintf("%d", i)
-		result, ok := decimalModeToOctal(i)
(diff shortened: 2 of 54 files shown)


base/util/merge_test.go
```diff
@@ -141,9 +142,11 @@ func TestMergeTranslatedConfigs(t *testing.T) {
 		},
 	}
 	for i, test := range tests {
-		c, ts := MergeTranslatedConfigs(test.parent, test.parentTranslations, test.child, test.childTranslations)
-		assert.Equal(t, test.merged, c, "#%d: bad config", i)
-		assert.Equal(t, test.mergedTranslations, ts, "#%d: bad translations", i)
+		t.Run(fmt.Sprintf("merge %d", i), func(t *testing.T) {
+			c, ts := MergeTranslatedConfigs(test.parent, test.parentTranslations, test.child, test.childTranslations)
+			assert.Equal(t, test.merged, c, "#%d: bad config", i)
```
bgilbert, line 147: If you're up for it, it'd be nice to remove all of the `#%d: ` stuff, since it's now redundant.

bgilbert (commented):
This is great, thank you!  A couple minor nits, plus a potentially large `sed` run.

There's one commit whose message isn't hard-wrapped.  Ideally commit messages would be kept to 72 characters.

Are there additional build flags we should be enabling for warnings?  I'm surprised the unused parameter didn't generate one.

Okeanos:
I'll fix the commit messages to be <=72 characters and see about the other changes later today.

> Are there additional build flags we should be enabling for warnings? I'm surprised the unused parameter didn't generate one.

I was surprised by that as you are about that particular detail – especially given that `go fmt` is usually very strict already. Personally, I used IDE code inspections to find most of these but there are projects such as [golangci-lint](https://golangci-lint.run) that would've probably found some of them as well. It gets tricky when considering the `os.TempDir` and subtest changes, though.

Okeanos:
I believe I addressed all previous review findings. Additionally the `#%d` within subtests should also be gone (yay for regex-replace) – someone who hasn't seen these changes multiple times (like me) should verify that, though.

jlebon:
Commits look great to me, though I've only skimmed the last one. Will let @bgilbert look at those and stamp the PR.

Wanted to say thank you for taking the time to hack on Butane and doing these changes. It's greatly appreciated.

bgilbert:
We do [run](https://github.com/coreos/butane/blob/main/.github/workflows/go.yml) golangci-lint in CI, but we may not be configuring it for maximum lintfulness.

bgilbert (approved):
This looks great; thank you very much!  A couple small nits, but in the interest of efficiency, I'll merge this now and address those in a followup.

bgilbert:
Followup in #313.

PR #315, opened 2022-01-25
docs: add information on PRs for the build process

Okeanos:
As discussed in and evident from
https://src.fedoraproject.org/rpms/butane/pull-request/5 the PR process
for non-packagers is not exactly obvious and requires additional local
setup to work.


bgilbert (approved):
Thanks!

PR #316, opened 2022-01-26
config/fcos/v1_5_exp: use array of extension instead of string

jmarrero:
Use []struct: https://github.com/coreos/butane/pull/304#discussion_r791945381

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -69,4 +69,7 @@ var (
 	ErrUserFieldSupport       = errors.New("fields other than \"name\" and \"ssh_authorized_keys\" are not supported in this spec version")
 	ErrUserNameSupport        = errors.New("users other than \"core\" are not supported in this spec version")
 	ErrKernelArgumentSupport  = errors.New("this field cannot be used for kernel arguments in this spec version; use openshift.kernel_arguments instead")
+
+	// Extensions
+	ErrExtensionNameRequired = errors.New("field \"name\" is required")
 )
--- a/config/fcos/v1_5_exp/schema.go
+++ b/config/fcos/v1_5_exp/schema.go
@@ -20,8 +20,8 @@ import (
 
 type Config struct {
 	base.Config `yaml:",inline"`
-	BootDevice  BootDevice `yaml:"boot_device"`
-	Extensions  Extensions `yaml:"extensions"`
+	BootDevice  BootDevice  `yaml:"boot_device"`
+	Extensions  []Extension `yaml:"extensions"`
 }
 
 type BootDevice struct {
@@ -40,4 +40,6 @@ type BootDeviceMirror struct {
 	Devices []string `yaml:"devices"`
 }
 
-type Extensions []string
+type Extension struct {
+	Name string `yaml:"name"`
+}
--- a/config/fcos/v1_5_exp/validate.go
+++ b/config/fcos/v1_5_exp/validate.go
@@ -39,3 +39,10 @@ func (m BootDeviceMirror) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (e Extension) Validate(c path.ContextPath) (r report.Report) {
+	if e.Name == "" {
+		r.AddOnError(c.Append("name"), common.ErrExtensionNameRequired)
+	}
+	return
+}
--- a/config/fcos/v1_5_exp/validate_test.go
+++ b/config/fcos/v1_5_exp/validate_test.go
@@ -187,3 +187,44 @@ func TestValidateBootDevice(t *testing.T) {
 		})
 	}
 }
+
+// TestValidateExtension tests extension validation
+func TestValidateExtension(t *testing.T) {
+	tests := []struct {
+		in      Extension
+		out     error
+		errPath path.ContextPath
+	}{
+		// No name field
+		{
+			Extension{},
+			common.ErrExtensionNameRequired,
+			path.New("yaml", "name"),
+		},
+		// complete config
+		{
+			Extension{
+				Name: "strace",
+			},
+			nil,
+			path.New("yaml"),
+		},
+		// empty string as name
+		{
+			Extension{
+				Name: "",
+			},
+			common.ErrExtensionNameRequired,
+			path.New("yaml", "name"),
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("validate %d", i), func(t *testing.T) {
+			actual := test.in.Validate(path.New("yaml"))
+			expected := report.Report{}
+			expected.AddOnError(test.errPath, test.out)
+			assert.Equal(t, expected, actual, "bad validation report")
+		})
+	}
+}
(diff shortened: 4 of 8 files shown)


cgwalters (commented):
Looks sane to me!

bgilbert (commented):
One nit; LGTM otherwise.

config/fcos/v1_5_exp/schema.go
```diff
@@ -40,4 +40,8 @@ type BootDeviceMirror struct {
 	Devices []string `yaml:"devices"`
 }
 
-type Extensions []string
+type Extensions []Extension
```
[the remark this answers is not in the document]
jmarrero: done

bgilbert:
Oh, also, spec docs need updating.

jmarrero:
> Oh, also, spec docs need updating.

Added the doc based on another list of objects I saw in the docs. Let me know if I missed anything.

docs/config-fcos-v1_5-exp.md
```diff
@@ -206,7 +206,8 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_threshold_** (int): sets the minimum number of pieces required to decrypt the device. Default is 1.
   * **_mirror_** (object): describes mirroring of the boot disk for fault tolerance.
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
-* **_extensions_** (list of strings): a list of packages to layer on top of the OS.
+* **_extensions_** (list of objects): the list of packages to layer on top of the OS.
+  * **name** (string): the name of the package.
```
bgilbert, line 210: This is correctly marked required here, but we're not actually enforcing that.  It's not _strictly_ necessary to do so (and it's unlikely that a user will forget the field until we have other fields in the struct) but would provide better UX than failing at runtime.  Could we add validation in `validate.go`, and a test for the validation?

docs/config-fcos-v1_5-exp.md
```diff
@@ -206,7 +206,8 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_threshold_** (int): sets the minimum number of pieces required to decrypt the device. Default is 1.
   * **_mirror_** (object): describes mirroring of the boot disk for fault tolerance.
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
-* **_extensions_** (list of strings): a list of packages to layer on top of the OS.
+* **_extensions_** (list of objects): the list of packages to layer on top of the OS.
```
bgilbert, line 209: Are these actually arbitrary package names, or are they a more restricted set of strings?

jmarrero: anything on the repos or rpm-ostree install allows

config/fcos/v1_5_exp/validate.go
```diff
@@ -39,3 +39,10 @@ func (m BootDeviceMirror) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (e Extension) Validate(c path.ContextPath) (r report.Report) {
+	if e.Name == "" {
+		r.AddOnError(c.Append("extensions"), common.ErrPackageNameRequired)
```
bgilbert, line 45: The path here is the field name relative to the function receiver:

```suggestion
		r.AddOnError(c.Append("name"), common.ErrPackageNameRequired)
```

config/fcos/v1_5_exp/validate_test.go
```diff
@@ -187,3 +187,44 @@ func TestValidateBootDevice(t *testing.T) {
...
+			common.ErrPackageNameRequired,
+			path.New("yaml", "extensions"),
+		},
+		// complete config
+		{
+			Extension{
+				Name: "strace",
+			},
+			nil,
+			path.New("yaml", "extensions"),
```
bgilbert, line 210: To avoid being misleading, let's just do this for nil error cases:

```suggestion
			path.New("yaml"),
```

config/common/errors.go
```diff
@@ -69,4 +69,7 @@ var (
 	ErrUserFieldSupport       = errors.New("fields other than \"name\" and \"ssh_authorized_keys\" are not supported in this spec version")
 	ErrUserNameSupport        = errors.New("users other than \"core\" are not supported in this spec version")
 	ErrKernelArgumentSupport  = errors.New("this field cannot be used for kernel arguments in this spec version; use openshift.kernel_arguments instead")
+
+	// Extensions
+	ErrPackageNameRequired = errors.New("field \"name\" is required")
```
bgilbert, line 74: To keep our terminology consistent, `ErrExtensionNameRequired`?

docs/config-fcos-v1_5-exp.md
bgilbert: Hmm, maybe I'm missing something here.  The existing MachineConfig `extensions` field supports only a restricted set of extension names, not arbitrary packages, and I had understood that this field was intended to be similar.  If this is intended for package installation, should we be calling it `packages` instead of `extensions`?

bgilbert: ^ cc @jlebon

jmarrero: It's similar as in it allows you to extend your OS. But this effort is part of the layering enhancement, as part of that we are trying to enable users to layer whatever they need. We are still having multiple conversation about the enhancement so naming could change. The idea of using Butane to declare the extensions/packages allow us to do something like: https://github.com/coreos/fcos-derivation-example/blob/main/Dockerfile but in a way that we can add more validation around it. A proof of concept of the effort is here: https://github.com/mkenigs/coreos-derive/tree/main/prototype#readme

cgwalters: For the MCO the fact that the available packages/extensions are lifecycle bound and tested with the cluster is an important aspect of things, IMO. FWIW this came from https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md

I can definitely see the argument for calling this `packages:` though.

(But I also have been trying for years now to create a distinction between "base image" and "addons/extensions", and so with this it's much more clear that one can't just do e.g. `podman-3.2.0` to override something from the base OS)

There's also the "why not both" solution of
```
extensions:
  packages:
    - name: usbguard
```

I don't feel *really* strongly on this TBH.

jlebon: The allowed set from the rpm-ostree side is really any "provides" which include e.g. package names, virtual provides, and even filepaths. Not sure if there's a good reason to try to be stricter here than what rpm-ostree supports.

I like the connotation of extensions to mean packages which really have to be on the host, and also like that we're leaving the door open for other kinds of extensions, so personally would leave it as is.

Minor suggestion, maybe just

```suggestion
* **_extensions_** (list of objects): the list of additional packages to be installed.
```

to be less implementation-y?

bgilbert (approved):

PR #325, opened 2022-02-14
WIP: adding support for multiarch container building

stobbsm:
- currently working with docker buildx/buildkit
- TODO: figure out multi arch container building in podman
- THIS IS NOT READY YET. But I want to have some comments before I proceed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
 FROM registry.fedoraproject.org/fedora:35 AS builder
+ARG BUILDARCH
+ENV BUILDARCH=${BUILDARCH}
 RUN dnf install -y golang git
 RUN mkdir /butane
 COPY . /butane
@@ -6,5 +8,7 @@ WORKDIR /butane
 RUN ./build_for_container
 
 FROM scratch
-COPY --from=builder /butane/bin/container/butane-x86_64-unknown-linux-gnu /usr/local/bin/butane
+ARG BUILDARCH
+ENV BUILDARCH=${BUILDARCH}
+COPY --from=builder /butane/bin/container/butane-${BUILDARCH}-unknown-linux-gnu /usr/local/bin/butan
 ENTRYPOINT ["/usr/local/bin/butane"]
--- a/build_for_container
+++ b/build_for_container
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+echo "Building for $BUILDARCH"
+
 set -euo pipefail
 
 export GO111MODULE=on
@@ -20,5 +22,5 @@ build_release() {
 }
 
 export GOOS=linux
-export GOARCH=amd64
-build_release x86_64-unknown-linux-gnu
+export GOARCH=$BUILDARCH
+build_release ${BUILDARCH}-unknown-linux-gnu


Dockerfile
```diff
@@ -1,10 +1,14 @@
...
 RUN mkdir /butane
 COPY . /butane
 WORKDIR /butane
 RUN ./build_for_container
 
 FROM scratch
-COPY --from=builder /butane/bin/container/butane-x86_64-unknown-linux-gnu /usr/local/bin/butane
+ARG BUILDARCH
+ENV BUILDARCH=${BUILDARCH}
+COPY --from=builder /butane/bin/container/butane-${BUILDARCH}-unknown-linux-gnu /usr/local/bin/butan
```
bgilbert, line 13: `/usr/local/bin/butane`

build_for_container
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+echo "Building for $BUILDARCH"
+
 set -euo pipefail
```
bgilbert, line 5: We usually initialize error handling first thing in the script, so let's move this above the `echo`.

Also, an environment variable is a little non-obvious, especially since `set -u` makes its existence required; let's just use a command-line argument instead (and print a usage message if not present).  That should be okay UX since this is only invoked for debugging or by the Dockerfile.

build_for_container
```diff
@@ -20,5 +22,5 @@ build_release() {
 }
 
 export GOOS=linux
-export GOARCH=amd64
-build_release x86_64-unknown-linux-gnu
+export GOARCH=$BUILDARCH
+build_release ${BUILDARCH}-unknown-linux-gnu
```
bgilbert, line 26: This will change the name to `amd64-*` but it doesn't really matter here.  In that case, since we're no longer using a valid GNU host triple, maybe let's drop the `-unknown-linux-gnu` suffix.

Dockerfile
```diff
@@ -1,10 +1,14 @@
...
 RUN dnf install -y golang git
 RUN mkdir /butane
 COPY . /butane
 WORKDIR /butane
 RUN ./build_for_container
 
 FROM scratch
-COPY --from=builder /butane/bin/container/butane-x86_64-unknown-linux-gnu /usr/local/bin/butane
+ARG BUILDARCH
+ENV BUILDARCH=${BUILDARCH}
```
bgilbert, line 12: Shouldn't be necessary, right?

bgilbert (commented):
Looks like a good start.  Quay won't successfully build this, but we knew we'd have to switch to other infrastructure.

Dockerfile
stobbsm: We still need to set the build arch, as go can build arm64 on amd64.

Going to need a fee days. Had emergency surgery.

bgilbert: Right, it's needed in the build stage, but I don't think it's needed in this stage.

Oof, hope you feel better soon.  There's no rush on this PR; take as much time as you need.

stobbsm: I see what you mean. Will take care of it as soon as possible.

Still need to figure out how to set the manifest to allow both amd64 and arm64.

bgilbert:
I had occasion to build a [GitHub Actions workflow](https://github.com/coreos/mkpasswd-container/blob/main/.github/workflows/containers.yml) that pushes a multi-arch container to Quay.  It's not a perfect match for the Butane case (we don't want to cross-build Butane, and can drop the README editing step) but it might be a good starting point for the CI job.

bgilbert:
I've made some improvements to the mkpasswd-container build code and moved it into a [GitHub Action](https://github.com/coreos/actions-lib/tree/main/build-container).  #333 adds that action to Butane CI, allowing us to ship a multi-arch container image.  Unfortunately we need to build arm64 in emulation until Buildah is updated in the GitHub Actions `ubuntu-latest` image, but the build performance is good enough for now.

Closing in favor of #333.  Thanks for the PR!

PR #327, opened 2022-02-22
docs/examples: use `mkpasswd` container

bgilbert:
RHEL 8 ships a different implementation of `mkpasswd` that doesn't support modern hashes.  For ease of use, just containerize `mkpasswd`.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -66,10 +66,10 @@ passwd:
       shell: /bin/bash
 ```
 
-To generate a secure password hash, use the `mkpasswd` command:
+To generate a secure password hash, use `mkpasswd`:
 
 ```
-$ mkpasswd --method=yescrypt
+$ podman run -ti --rm quay.io/coreos/mkpasswd --method=yescrypt
 Password:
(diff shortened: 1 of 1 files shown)


docs/examples.md
```diff
@@ -74,6 +74,14 @@ Password:
 $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
 ```
 
+If you don't have `mkpasswd`, or your copy doesn't support `--method`, you can run `mkpasswd` in a container:
+
+```
+$ podman run -ti --rm quay.io/coreos/mkpasswd --method=yescrypt
```
lucab, line 80: I'd be in favor of directly replacing the straight `mkpasswd` example above and only keeping this new mkpasswd-in-podman snippet here.

That said, I couldn't see where this container is coming from. It would be useful to have a breadcrumb to its source repo, perhaps directly in this doc.

lucab: That's more than enough, thanks!

lucab (approved):

PR #338, opened 2022-04-08
Dockerfile: Use a full base image

cgwalters:
I'm trying to use butane in a `Dockerfile` as part of a multi-stage
build, and not having anything except the binary in the container
breaks that since `Dockerfile` wants to invoke things via `/bin/sh`.

Use ubi8 as a base.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,6 @@ COPY . /butane
 WORKDIR /butane
 RUN ./build_for_container
 
-FROM scratch
+FROM registry.fedoraproject.org/fedora-minimal:35
 COPY --from=builder /butane/bin/container/butane /usr/local/bin/butane
 ENTRYPOINT ["/usr/local/bin/butane"]


bgilbert:
Hmm, yeah, that makes sense.  Let's use `registry.fedoraproject.org/fedora:35` or `registry.fedoraproject.org/fedora-minimal:35` as a base though.  We usually don't use the UBI in our upstreams.

Also, is the path change important for your use case?  I'm wondering if that risks a compat break somewhere.

cgwalters:
> We usually don't use the UBI in our upstreams.

There's no restrictions on UBI by design; a specific advantage over fedora as a host here is we don't need to churn this as often.

But we are unlikely to be consistent on this, and it doesn't matter much to me so...switched to fedora.

cgwalters:
> Also, is the path change important for your use case? 

Nah, I just thought it was cleaner.  Reverted.

bgilbert:
> There's no restrictions on UBI by design; a specific advantage over fedora as a host here is we don't need to churn this as often.

Yeah, I know.  But I'd feel weird about a community project using a non-community base image.  Thanks for making the switch.

bgilbert (approved):

PR #339, opened 2022-04-11
v1_5_exp: Add sugar to create user.cfg

saqibali-2k:
We want to add sugar to butane which will allow users to indirectly
modify the GRUB configuration. The sugar added here will abstract
the mounting of `/boot` and will allow users to create `/boot/grub2/user.cfg`
which is sourced by `grub.cfg`.


base/v0_5_exp/schema.go
```diff
@@ -84,6 +84,11 @@ type FilesystemOption string
 
 type Group string
 
+type Grub struct {
+	Inline *string `yaml:"inline"`
+	Local  *string `yaml:"local"`
```
bgilbert, line 89: The user could just declare the `/boot` filesystem and the `user.cfg` contents directly, right?  If all we want to do is let users provide an arbitrary `user.cfg`, we'd normally document an example config, rather than implementing sugar.

I do think sugar would be useful here, but it should be at a higher semantic level.  Rather than exposing GRUB config syntax, we could provide fields for specific configuration knobs, similar to the existing `PasswdUser` struct.  That also helps discourage users from trying to arbitrarily change the bootloader configuration, which we surely don't want to support.  Is there a specific list of features/commands we want to enable?

bgilbert (commented):
Let's mention https://github.com/coreos/fedora-coreos-tracker/issues/134 in the commit message.

base/v0_5_exp/schema.go
saqibali-2k: The only one we need to enable currently is GRUB password support. Other than that, I haven't tinkered much with GRUB configurations so I'm not sure what some nice-to-haves would be.

config/fcos/v1_5_exp/translate.go
```diff
@@ -350,3 +355,77 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+	}
+
+	// create boot filesystem
+	bootFilesystem := types.Filesystem{
+		Device: "/dev/disk/by-label/boot",
+		Format: util.StrToPtr("ext4"),
+		Path:   util.StrToPtr("/boot"),
+	}
+
+	// ts.AddFromCommonSource(yamlPath, path.New("json", "storage", "filesystems", len(rendered.Storage.Filesystems)), bootFilesystem)
```
bgilbert, line 377: Some stray commented code.

config/fcos/v1_5_exp/validate.go
```diff
@@ -46,3 +47,10 @@ func (e Extension) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
+	if util.NotEmpty(user.Key) && !util.NotEmpty(user.User) {
```
bgilbert, line 52: Is there any reason to accept GrubUsers with missing usernames?

bgilbert (commented):
Thanks for handling this!  There are some details, but this has the right general idea. :+1:

config/fcos/v1_5_exp/validate.go
saqibali-2k: You're right, there isn't. Will fix the condition :+1:

config/fcos/v1_5_exp/translate.go
```diff
@@ -350,3 +355,77 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+		file.Contents.Compression = util.StrToPtr("gzip")
+	}
+	// ts.AddFromCommonSource(yamlPath, path.New("json", "storage", "files", len(rendered.Storage.Files)), file)
+
+	rendered.Storage.Files = append(rendered.Storage.Files, file)
+	ts.AddFromCommonSource(yamlPath, path.New("json", "storage"), rendered.Storage)
+	return rendered, ts, r
+}
+
+func (gb Grub) BuildConfig() string {
```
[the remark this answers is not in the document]
saqibali-2k: Do we still want to use the builder pattern here? I felt that it would make future additions to the functionality easier.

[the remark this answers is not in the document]
bgilbert: It's a reasonable pattern, but at present we don't generally use it in Butane.  For consistency, let's avoid for now.

bgilbert:
As I mentioned in https://github.com/coreos/butane/pull/339#discussion_r847646381, I think we should actively discourage users from making arbitrary changes to their GRUB configuration.  By limiting the available sugar and not documenting how to make arbitrary changes, we can make `user.cfg` a semi-private API between FCOS and Butane.  We can always add additional sugar later, including support for arbitrary config fragments if we really want to.

saqibali-2k:
Example Config:
```yaml
variant: fcos
version: 1.5.0-experimental
grub:
  users:
    - user: root
      key: grub.pbkdf2.sha512.10000.874A958E526409...
```

saqibali-2k:
Rebased `baseutil.MakeDataURL` changes and pushed documentation updates.

config/fcos/v1_5_exp/translate.go
```diff
@@ -348,3 +353,70 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+
+	for _, user := range gb.Users {
+		// We have already validated that user.Name is non-empty
+		allUsers += user.Name + " "
+		if util.NotEmpty(user.PasswordHash) {
+			// Command for setting users password
+			cmdList += fmt.Sprintf("password_pbkdf2 %s %s\n", user.Name, *user.PasswordHash)
+		}
+	}
+	superUsrCmd := fmt.Sprintf("set superusers=\"%s\"\n", allUsers)
```
bgilbert, line 417: Tiny nit: the missing `e` in `user` is oddly inconsistent.

config/fcos/v1_5_exp/translate.go
```diff
@@ -348,3 +353,70 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+
+	contents += superUsrCmd
+	contents += cmdList
+	return contents
```
bgilbert, line 421: Could just be:
```suggestion
	return "# Generated by Butane\n\n" + superUserCmd + strings.Join(cmds, "\n")
```

config/fcos/v1_5_exp/translate.go
```diff
@@ -348,3 +353,70 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+	ts := translate.NewTranslationSet("yaml", "json")
+	var r report.Report
+	yamlPath := path.New("yaml", "grub")
+	if len(c.Grub.Users) == 0 {
+		// No users
+		return rendered, ts, r
+	}
+
+	// create boot filesystem
+	bootFilesystem := types.Filesystem{
```
bgilbert, line 369: Nit: unnecessary temporary.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1487,3 +1487,80 @@ func TestTranslateExtensions(t *testing.T) {
 		})
 	}
 }
+
+// TestTranslateExtensions tests translating the Butane config extensions section.
```
bgilbert, line 1491: Stale comment.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1487,3 +1487,80 @@ func TestTranslateExtensions(t *testing.T) {
...
+func TestTranslateGrub(t *testing.T) {
+	tests := []struct {
+		in         Config
+		out        types.Config
+		exceptions []translate.Translation
+		report     report.Report
+	}{
+		// config with 1 user
```
bgilbert, line 1499: Should also test with multiple users, and a user without a password hash.

config/fcos/v1_5_exp/validate_test.go
```diff
@@ -228,3 +228,39 @@ func TestValidateExtension(t *testing.T) {
...
+		// username is not specified but password is (invalid)
+		{
+			in: GrubUser{
+				Name:         "",
+				PasswordHash: util.StrToPtr("pkcs5-pass"),
+			},
+			out:     common.ErrGrubUserNameNotSpecified,
+			errPath: path.New("yaml", "grub", "users"),
+		},
+	}
```
bgilbert, line 256: Username is specified and password is missing?

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 * **_extensions_** (list of objects): the list of additional packages to be installed.
   * **name** (string): the name of the package.
+* **_grub_** (object): describes configuration changes for GRUB
+  * **_users_** (list of objects): the list of GRUB super users.
```
bgilbert, line 212: `superusers`

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
...
+    * **name** (string): the user name.
+    * **passwordHash** (string): the pbkdf2 password hash
```
bgilbert, line 214: `PBKDF2`

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 * **_extensions_** (list of objects): the list of additional packages to be installed.
   * **name** (string): the name of the package.
+* **_grub_** (object): describes configuration changes for GRUB
```
bgilbert, line 211: Hmm, maybe something like `describes the desired GRUB bootloader configuration.`?  `Changes` is imperative; we try to use declarative language in the spec.

bgilbert (commented):
Code generally looks good!  Just a pile of small cleanups. :broom:

config/fcos/v1_5_exp/validate_test.go
saqibali-2k: That would just be setting a superuser without a password. Is that something that we want to prohibit?

config/fcos/v1_5_exp/translate.go
```diff
@@ -348,3 +354,67 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+	ts := translate.NewTranslationSet("yaml", "json")
+	var r report.Report
+	yamlPath := path.New("yaml", "grub")
+	if len(c.Grub.Users) == 0 {
+		// No users
+		return rendered, ts, r
+	}
+
+	// create boot filesystem
+	// bootFilesystem :=
```
bgilbert, line 369: Stray.

config/fcos/v1_5_exp/translate.go
```diff
@@ -348,3 +354,67 @@ func (c Config) processPackages(options common.TranslateOptions) (types.Config,
...
+		// We have already validated that user.Name is non-empty
+		allUsers = append(allUsers, user.Name)
+		if util.NotEmpty(user.PasswordHash) {
+			// Command for setting users password
+			contents = append(contents, fmt.Sprintf("password_pbkdf2 %s %s\n", user.Name, *user.PasswordHash))
+		}
+	}
+	superUserCmd := fmt.Sprintf("set superusers=\"%s\"\n", strings.Join(allUsers, " "))
+	// Slice will always be of atleast size 2 since we initialized it as such
+	contents[1] = superUserCmd
```
bgilbert, line 418: This works but is a bit trickier than necessary.  Instead of `contents`, I'd be inclined to have a slice of password entries, and only build up the contents at the end of the function.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1487,3 +1487,195 @@ func TestTranslateExtensions(t *testing.T) {
...
+			Config{
+				Grub: Grub{
+					Users: []GrubUser{
+						{
+							Name:         "root1",
+							PasswordHash: util.StrToPtr("grub.pbkdf2.sha512.10000.874A958E526409..."),
+						},
+						{
+							Name:         "root2",
+							PasswordHash: util.StrToPtr("grub.pbkdf2.sha512.10000.874A958E526409..."),
```
bgilbert, line 1566: It's better to set different hashes, so the tests will notice if we get them mixed up somehow.

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 * **_extensions_** (list of objects): the list of additional packages to be installed.
   * **name** (string): the name of the package.
+* **_grub_** (object): describes the desired GRUB bootloader configuration
```
bgilbert, line 211: We use a trailing period for each entry.

docs/examples.md
```diff
@@ -372,6 +372,20 @@ systemd:
         [Install]
         WantedBy=multi-user.target
 ```
+## GRUB password
+
+This example adds a superuser to GRUB and sets a password. Users without the given username
+and password will not be able to access GRUB command-line, modify kernel command-line arguments, or boot non-default OSTree deployments.
```
bgilbert, line 378: The first instance of `command-line` is a compound noun and should be `command line`.

config/fcos/v1_5_exp/validate_test.go
bgilbert: We currently don't prohibit it, so we should test for it.  But, yeah, it doesn't seem to make sense to allow adding a superuser without a password.  It's safest to forbid it for now, since we can always relax the restriction later.

bgilbert:
Currently, if a parent Butane config and a child Butane config both specify `grub.users`, then when Ignition merges the resulting Ignition configs at runtime, the parent users will be completely overwritten by the child.  This is surprising, since it violates the usual config-merging semantics of replacing existing list entries and appending new ones.

In principle we could provide the usual semantics by having the sugar use `storage.files.append` instead of `storage.files.contents`, and using `set superusers="$superusers ..."`.  A couple concerns though:

- This assumes that GRUB overwrites an existing password if it sees a second `password_pbkdf2` command for a given user.  @saqibali-2k, could you verify this?
- This wouldn't work in `openshift` because the MCO doesn't support `append`.  I'm not sure that should stop us though.  The `openshift` variant can always have special-case code to look for `user.cfg` and move its `append` directive to `contents`.  @jlebon, any thoughts?

saqibali-2k:
> * This assumes that GRUB overwrites an existing password if it sees a second `password_pbkdf2` command for a given user.  @saqibali-2k, could you verify this?

Verified this: the old password does not work after a second `password_pbkdf2` command is given.

bgilbert:
(And I assume it's okay to list the same user twice in `superusers`.)

saqibali-2k:
> (And I assume it's okay to list the same user twice in `superusers`.)

This is true, but this resets any previously defined password for that user. So in practice if no new password is set, we just get a username where the password has been thrown away.

bgilbert:
To be clear, are you saying that `set superusers="foo"` deletes any previous password for `foo`?

I think that's actually okay.  As long as we define `password_hash` as required, we won't add a user to `superusers` unless we're also setting a password for it.

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 * **_extensions_** (list of objects): the list of additional packages to be installed.
   * **name** (string): the name of the package.
+* **_grub_** (object): describes the desired GRUB bootloader configuration.
+  * **_users_** (list of objects): the list of GRUB superusers.
+    * **name** (string): the user name.
+    * **passwordHash** (string): the PBKDF2 password hash.
```
bgilbert, line 214: `password_hash`

config/fcos/v1_5_exp/validate_test.go
```diff
@@ -228,3 +228,47 @@ func TestValidateExtension(t *testing.T) {
...
+				Name: "name",
+			},
+			out:     common.ErrGrubPasswordNotSpecified,
+			errPath: path.New("yaml", "grub", "password_hash"),
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("validate %d", i), func(t *testing.T) {
+			actual := test.in.Validate(path.New("yaml", "grub"))
```
bgilbert, line 268: This is a bit confusing, relative to our practices in other tests.  Let's just use `path.New("yaml")` for consistency.

bgilbert (commented):
Thanks for switching to `append`.  This should currently fail validation in the `openshift` experimental spec.  Want to add some translation code there to find the `user.cfg` entry and convert its `append` to `contents`?

docs/config-fcos-v1_5-exp.md
```diff
@@ -208,6 +208,11 @@ The Fedora CoreOS configuration is a YAML document conforming to the following s
     * **_devices_** (list of strings): the list of whole-disk devices (not partitions) to include in the disk array, referenced by their absolute path. At least two devices must be specified.
 * **_extensions_** (list of objects): the list of additional packages to be installed.
   * **name** (string): the name of the package.
+* **_grub_** (object): describes the desired GRUB bootloader configuration.
+  * **_users_** (list of objects): the list of GRUB superusers.
+    * **name** (string): the user name.
+    * **password_hash** (string): the PBKDF2 password hash.
+
```
bgilbert, line 215: Nit: stray newline.

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +296,24 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
 	}
 	return cutil.TranslateReportPaths(r, ts)
 }
+
+// fcos v1.5.0-exp generates a user.cfg file using append. Change this to contents
+func handleUserGrubCfg(config *types.Config, ts *translate.TranslationSet) {
```
bgilbert, line 301: Maybe `translateUserGrubCfg`?

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +296,24 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+						Source:      file.Append[0].Source,
+						Compression: file.Append[0].Compression,
+					},
+				},
+			}
+			ts.AddFromCommonSource(path.New("yaml", "grub", "users"), path.New("json", "storage", "files", i), config.Storage.Files[i])
+			return
+		}
+	}
+}
```
bgilbert, line 319: Let's add a test for this.

bgilbert (commented):
Thanks for making the additional changes!

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +297,28 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+				translatedPath := path.New("json", "storage", "files", i, "contents")
+				config.Storage.Files[i].FileEmbedded1.Contents.Source = file.Append[0].Source
+				newMappings.AddTranslation(path.New("json", "storage", "files", i, "append", 0, "source"), translatedPath.Append("source"))
+				config.Storage.Files[i].FileEmbedded1.Contents.Compression = file.Append[0].Compression
+				newMappings.AddTranslation(path.New("json", "storage", "files", i, "append", 0, "compression"), translatedPath.Append("compression"))
+				config.Storage.Files[i].FileEmbedded1.Append = nil
+
+				newMappings.AddTranslation(path.New("json", "storage", "files", i, "append", 0), translatedPath)
+			} else {
+				panic(fmt.Errorf("translating user.cfg: expected one append object but instead %v were found", len(config.Storage.Files)))
```
bgilbert, line 317: It's possible for a user to have (improperly) done this by hand:

```yaml
variant: openshift
version: 4.12.0-experimental
metadata:
  name: foo
  labels:
storage:
  files:
    - path: /boot/grub2/user.cfg
      append:
        - inline: foo
        - inline: bar
```

so we can't panic here because that's reserved for programmer errors.  The existing validation code will notice and reject the `append` section later, so I think all we have to do here is return early.

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +297,28 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
 	return cutil.TranslateReportPaths(r, ts)
 }
+
+// fcos v1.5.0-exp generates a user.cfg file using append; however, OpenShift config
+// does not support append (since MCO does not support it). Let change the file to use contents
+func translateUserGrubCfg(config *types.Config, ts *translate.TranslationSet) translate.TranslationSet {
+	newMappings := translate.NewTranslationSet("json", "json")
+	for i, file := range config.Storage.Files {
+		if file.Path == "/boot/grub2/user.cfg" {
+			if len(file.Append) == 1 {
```
bgilbert, line 307: The sad path should error out early so the happy path can avoid the extra indent.

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +296,24 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+				Node: types.Node{
+					Path: "/boot/grub2/user.cfg",
+				},
+				FileEmbedded1: types.FileEmbedded1{
+					Contents: types.Resource{
+						Source:      file.Append[0].Source,
+						Compression: file.Append[0].Compression,
+					},
+				},
+			}
```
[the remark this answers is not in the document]
bgilbert: We're still assuming that the fields we're ignoring are, in fact, empty.  The user might have put something in them.

Rather than checking the other fields one at a time, it might be cleaner to assign the entirety of `Append[0]` to `Contents` and update the TranslationSet accordingly.  We don't have a helper to do the latter right now, but `AddFromCommonSource` is close.  We could add a variant of `AddFromCommonSource` that appends subpaths to both the `From` and `To` side, instead of just the `To` side.

config/openshift/v4_12_exp/translate_test.go
```diff
@@ -271,6 +271,87 @@ func TestTranslateConfig(t *testing.T) {
...
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "filesystems")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "filesystems", 0)},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "filesystems", 0, "path")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "filesystems", 0, "device")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "filesystems", 0, "format")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "files")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "files", 0)},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "files", 0, "path")},
+				{path.New("yaml", "grub", "users"), path.New("json", "spec", "config", "storage", "files", 0, "append")},
```
bgilbert, line 349: We should add a comment (here and/or in the code above) saying why this entry exists.

config/openshift/v4_12_exp/translate.go
saqibali-2k: Sorry where are we avoiding the extra indent?

bgilbert: Starting on the next line.

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +296,25 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+		if file.Path == "/boot/grub2/user.cfg" {
+			if len(file.Append) != 1 {
+				// The number of append objects was different from expected, this file
+				// was created by the user and not via butane GRUB sugar
+				return *ts
+			}
+			translatedPath := path.New("json", "storage", "files", i, "contents")
+			config.Storage.Files[i].FileEmbedded1.Contents = file.Append[0]
+			config.Storage.Files[i].FileEmbedded1.Append = nil
+			newMappings.AddFromCommonObject(path.New("json", "storage", "files", i, "append", 0), file.Append[0], translatedPath, file.Contents)
```
bgilbert, line 314: This doesn't affect the functionality because the field contents don't actually matter here, but: `file.Contents` is an empty struct.  It might be clearer to pass the populated copy of the struct.

Also, it's a bit odd that one of the paths is declared beforehand and the other one isn't.

bgilbert (commented):
Almost there; just some small fixes.

translate/set.go
```diff
@@ -105,6 +105,27 @@ func (ts TranslationSet) AddFromCommonSource(common path.ContextPath, toPrefix p
 	ts.AddTranslation(common, toPrefix)
 }
 
+// AddFromCommonObject adds translations for all of the paths in to from a matching path in from.
+// This is useful if both from and to have the same type (and does not make sense otherwise).
+func (ts TranslationSet) AddFromCommonObject(fromPrefix path.ContextPath, from interface{}, toPrefix path.ContextPath, to interface{}) {
```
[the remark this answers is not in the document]
saqibali-2k: I don't think we need it for retrieving the paths, but I think its useful for performing sanity checks:
`
vFrom.Type() != vTo.Type()
` and `
len(vPathsFrom) != len(vPathsTo)
`.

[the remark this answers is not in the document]
bgilbert: I'd prefer that we not.  By taking two objects, we're implying that the function meaningfully operates on both of them, when really we're just validating semantics that aren't really our responsibility.  (None of the existing `TranslationSet` methods validate that a newly created path matches a real struct field.)  The object is really just being used as a template for creating subpaths under the two specified roots.

Also, in the typical case that we're translating from `yaml` to `json`, the two structs could have identical fields but be of different types.

config/openshift/v4_12_exp/translate.go
```diff
@@ -295,3 +296,26 @@ func validateMCOSupport(mc result.MachineConfig, ts translate.TranslationSet) re
...
+	newMappings := translate.NewTranslationSet("json", "json")
+	for i, file := range config.Storage.Files {
+		if file.Path == "/boot/grub2/user.cfg" {
+			if len(file.Append) != 1 {
+				// The number of append objects was different from expected, this file
+				// was created by the user and not via butane GRUB sugar
+				return *ts
+			}
+			translatedPath := path.New("json", "storage", "files", i, "contents")
+			fromPath := path.New("json", "storage", "files", i, "append", 0)
```
bgilbert, line 312: _Tiny_ nit: these would be clearer in reverse order.

translate/set.go
```diff
@@ -105,6 +105,19 @@ func (ts TranslationSet) AddFromCommonSource(common path.ContextPath, toPrefix p
 	ts.AddTranslation(common, toPrefix)
 }
 
+// AddFromCommonObject adds translations for all of the paths in to from a matching path in from.
+// This is useful if both from and to have the same type (and does not make sense otherwise).
+func (ts TranslationSet) AddFromCommonObject(fromPrefix path.ContextPath, toPrefix path.ContextPath, to interface{}) {
+	vTo := reflect.ValueOf(to)
+	vPathsTo := prefixPaths(getAllPaths(vTo, ts.ToTag, true), toPrefix.Path...)
+	vPathsFrom := prefixPaths(getAllPaths(vTo, ts.FromTag, true), fromPrefix.Path...)
```
bgilbert, line 113: This is probably okay as-is, but there's an unnecessary implicit assumption that `getAllPaths()` returns paths in a consistent order.  Possible cleanup: call `getAllPaths` once, drop the `prefixPaths` calls, and then loop over paths:

```go
ts.AddTranslation(prefixPath(path, fromPrefix.Path...), prefixPath(path, toPrefix.Path...))
```

translate/set.go
```diff
@@ -105,6 +105,19 @@ func (ts TranslationSet) AddFromCommonSource(common path.ContextPath, toPrefix p
 	ts.AddTranslation(common, toPrefix)
 }
 
+// AddFromCommonObject adds translations for all of the paths in to from a matching path in from.
+// This is useful if both from and to have the same type (and does not make sense otherwise).
```
bgilbert, line 109: This is stale now.  (As is the commit message.)

translate/set.go
bgilbert: Followup in #357.

travier:
Awesome!

Issue #332, opened 2022-03-10
Automatically-compressed resources interact badly with config merging


cgwalters:
I'm working on updating the MCO to use the latest butane instead of fcct in https://github.com/openshift/machine-config-operator/pull/2976

There were a few minor bugs on the MCO side this unveiled, but this one looks like a clear Ignition bug.  Basically in the MCO we heavily do config merging.

Butane now uses compression by default (which is fine; but I had to fix up various places in the MCO to correctly handle this).  But, Ignition's native config merging seems to incorrectly carry the compression flag forward from old files when trying to override with an uncompressed file:

```
diff --git a/pkg/controller/common/helpers_test.go b/pkg/controller/common/helpers_test.go
index 8934b7d98..c667bbc33 100644
--- a/pkg/controller/common/helpers_test.go
+++ b/pkg/controller/common/helpers_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/clarketm/json"
 	ign2types "github.com/coreos/ignition/config/v2_2/types"
+	ign3 "github.com/coreos/ignition/v2/config/v3_2"
 	ign3types "github.com/coreos/ignition/v2/config/v3_2/types"
 	validate3 "github.com/coreos/ignition/v2/config/validate"
 	"github.com/stretchr/testify/assert"
@@ -135,6 +136,30 @@ func TestValidateIgnition(t *testing.T) {
 	require.NotNil(t, isValid3)
 }
 
+func TestIgnitionMergeCompressed(t *testing.T) {
+	testIgn3Config := ign3types.Config{}
+	testIgn3Config.Ignition.Version = "3.2.0"
+	mode := 420
+	testfiledata := "data:;base64,H4sIAAAAAAAAA0vLz+cCAKhlMn4EAAAA"
+	compression := "gzip"
+	tempFile := ign3types.File{Node: ign3types.Node{Path: "/etc/testfileconfig"},
+		FileEmbedded1: ign3types.FileEmbedded1{Contents: ign3types.Resource{Source: &testfiledata, Compression: &compression}, Mode: &mode}}
+	testIgn3Config.Storage.Files = append(testIgn3Config.Storage.Files, tempFile)
+
+	testIgn3Config2 := ign3types.Config{}
+	testIgn3Config2.Ignition.Version = "3.2.0"
+	testfiledata = "data:hello world"
+	tempFile = ign3types.File{Node: ign3types.Node{Path: "/etc/testfileconfig"},
[...]

PR #341, opened 2022-04-21
base/*: fix uncompressed inline child resource with compressed parent

bgilbert:
When desugaring inline/local resource contents, the current rules are:

1. If the input config specifies `compression: gzip`, we pass through the `compression` field unchanged and refuse to compress again.  This is useful for local resources that are stored precompressed alongside the config.
2. If we decide to compress, we set the `compression` field to `gzip`.
3. If we decide not to compress, we do not set the `compression` field, ostensibly so precompressed resources in child configs can inherit the `compression` field from the parent.

But case 3 doesn't really work.  Butane doesn't know that the child resource is precompressed, so is allowed to compress it again, and only the compressibility heuristics (which technically are not contractual) prevent it from doing so and producing invalid output.  (There's no way to represent nested compression, so the file would be written to the provisioned system compressed.)  And if the parent resource is not formally precompressed, but is actually just auto-compressed by Butane, a change to the parent's contents or to the Butane heuristics could cause the parent to become uncompressed in the future and break the merged config.

And case 3 creates a problem for an inline/local child resource overriding an inline/local parent.  If Butane opts to compress the parent but not the child, the merged config will have uncompressed contents but inherit `compression: gzip` from the parent, causing a provisioning failure.  And if the parent/child contents change over time, this problem may suddenly appear in the future.

Fix this by changing rule 3.  When desugaring a non-precompressed inline/local resource, always set the `compression` field to either "" or `gzip`.  This intuitively makes sense: only we know whether we decided to compress the resource, so we need to say so.

This involves an API change to `MakeDataURL()`.  We now return a pointer to the selected `compression` value, or `nil` if we're passing through the `compression` value from the input (case 1).  In the latter case, the caller should record a 1:1 translation for the `compression` field, rather than attributing its value to the `inline`/`local` input field.

Fixes https://github.com/coreos/butane/issues/332.


base/util/url.go
```diff
@@ -53,10 +65,10 @@ func MakeDataURL(contents []byte, currentCompression *string, allowCompression b
 			return
 		}
 		gz := ";base64," + base64.StdEncoding.EncodeToString(buf.Bytes())
-		// Account for space needed by "compression": "gzip".
-		if len(gz)+25 < len(opaque) {
+		// Account for space needed by the "gzip" compression value
+		if len(gz)+4 < len(opaque) {
```
cgwalters, line 69: I would make this `len(gz)+len("gzip")` or so...the compiler should constant fold hopefully, and it makes it more robust to future changes.

base/util/url.go
```diff
@@ -24,9 +24,21 @@ import (
...
+	// contents were compressed by the user; we can pick a data URL
+	// encoding but we can't compress again.  In this case we set our
+	// compression return value to null so the caller knows not to
+	// record a translation from input contents to output compression.
+	// In all other cases, we need to explicitly set the compression
+	// field to avoid a child config inheriting an incorrect compression
+	// setting from the parent.
+	if util.NilOrEmpty(currentCompression) {
+		compression = util.StrToPtr("")
+	}
```
cgwalters, line 40: I had to read this comment and code block twice.  The comment starts out talking about a condition, then talks about the "all other cases" - and the code implements the "all other cases".  I kept expecting it to implement what the *start* of the comment was describing.

Maybe:

```
If the config has an empty/nil compression field, we need to 
explicitly set the compression field to avoid a child config
inheriting an incorrect compression setting from the parent.

Otherwise, If the config has a non-empty compression field, then the
contents were compressed by the user; we can pick a data URL
encoding but we can't compress again.  In this case we set our
compression return value to null so the caller knows not to
record a translation from input contents to output compression.
```

maybe?

cgwalters (approved):
I only gave this a somewhat superficial look but seems sane to me!

bgilbert:
Updated.

base/v0_2/translate_test.go
```diff
@@ -196,7 +196,8 @@ func TestTranslateFile(t *testing.T) {
 							},
 						},
 						{
-							Source: util.StrToPtr("data:,file%20contents%0A"),
+							Source:      util.StrToPtr("data:,file%20contents%0A"),
+							Compression: util.StrToPtr(""),
```
cgwalters, line 200: Do we actually need to change all the tests to explicitly specify a `""` compression?  Shouldn't it be equivalent to `nil`?

jlebon (approved):
LGTM overall. Do we need to surface this somehow also in the Ignition spec docs for the Ignition-level version of the config merge issue?

cgwalters:
OK right, while this seems like a useful fix I don't think it will actually help the MCO today, see https://github.com/openshift/machine-config-operator/pull/3126#issuecomment-1113650288
which I guess we knew from https://github.com/coreos/butane/issues/332#issuecomment-1070861759
anyways.

base/v0_2/translate_test.go
bgilbert: Right, the whole point of this PR is to emit `""` rather than `nil` because they're semantically distinct for config merging.

bgilbert:
> Do we need to surface this somehow also in the Ignition spec docs for the Ignition-level version of the config merge issue?

We [describe the field merging semantics](https://coreos.github.io/ignition/operator-notes/#child-configs-take-precedence-when-specified) in Ignition operator notes, so I don't think so?  The issue here was that Butane wasn't taking those semantics into account.

base/v0_2/translate_test.go
cgwalters: > Right, the whole point of this PR is to emit "" rather than nil because they're semantically distinct for config merging.

Yeah sorry, it is obvious when stated that way.  I don't know why I find this so confusing.

bgilbert:
Filed https://github.com/coreos/ignition/issues/1357 for better config merging docs, including an explicit callout of overriding the `compression` field.

PR #344, opened 2022-05-07
github: enable Dependabot

bgilbert:
We have a lot of vendored dependencies, which we currently don't update.  Let's fix that.  We should have enough CI to pick up substantial regressions.

Hold until after release.


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: daily
```
jlebon, line 6: In rpm-ostree at least, we do `weekly` to have some sort of batching at the notifications level. Though here we don't really have that many deps so many it's fine.

jlebon (approved):

.github/dependabot.yml
bgilbert: That's a good point.  We're doing `daily` in the Rust projects and the paperwork load is honestly a bit high.  I'll PR to update everything.

bgilbert:
Reduced update frequency to weekly, to reduce paperwork.

bgilbert:
And updated the open PR count to 10, and configured labels.

Okeanos:
Btw. you can add a few other things such as Docker and GitHub Actions in here as well:

```yml
  - package-ecosystem: github-actions
    directory: "/"
    schedule:
      interval: "weekly"
    open-pull-requests-limit: 10
  - package-ecosystem: docker
    directory: "/"
    schedule:
      interval: "weekly"
    open-pull-requests-limit: 10
```

PR #350, opened 2022-05-20
translate: add `TranslationSet.Map()` to further translate `To` fields

bgilbert:
In special circumstances, postprocessing code may need to rearrange fields already translated to output space, in which case it needs to update the corresponding translations.  Provide a helper function to do so.

For https://github.com/coreos/butane/pull/339.

### DIFF
--- a/translate/set.go
+++ b/translate/set.go
@@ -160,6 +160,24 @@ OUTER:
 	return ret
 }
 
+// Map returns a new TranslationSet with To translation paths further
+// translated through mappings.  Translations not listed in mappings are
+// copied unmodified.
+func (ts TranslationSet) Map(mappings TranslationSet) TranslationSet {
+	if mappings.FromTag != ts.ToTag || mappings.ToTag != ts.ToTag {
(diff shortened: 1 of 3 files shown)


translate/set.go
```diff
@@ -160,6 +160,24 @@ OUTER:
 	return ret
 }
 
+// Map returns a new TranslationSet with To translation paths further
+// translated through mappings.  Translations not listed in mappings are
+// copied unmodified.
+func (ts TranslationSet) Map(mappings TranslationSet) TranslationSet {
+	if mappings.FromTag != ts.ToTag || mappings.ToTag != ts.ToTag {
```
saqibali-2k, line 167: When would we expect `mappings.ToTag == ts.ToTag` to be true? We use `mapping.From.String()` as keys for `ret.Set`, so wouldn't `mappings.FromTag == ts.ToTag` always need to be true? Thanks for any corrections

saqibali-2k (approved):
Thanks for adding this. LGTM, just had a question for my understanding.

translate/set.go
bgilbert: Yes, both should be true.  Concretely, `ts` will typically map `yaml` -> `json`, and `Map` performs mappings in output space, so `mappings` should be `json` -> `json`.

Issue #371, opened 2022-07-11
Consider Switching to type aliasing to reduce code duplication


prestist:
# Type aliasing suggestion
While evaluating an [issue in Ignition](https://github.com/coreos/ignition/issues/1397), I noticed the tendency to duplicate the schema's for each version. This was done to allow for adjusting a schema's individual attributes separately per schema version. However, it seems that in practice defined structs do not change per schema version, and tend to stay in place with a few exceptions. Furthermore with a larger and larger list of schemas we find ourselves duplicating patches that go into each schema, and then having to duplicate the unit tests for those changes. You can see a working example of this problem [here](https://github.com/coreos/ignition/pull/1412/commits/56ed15a6cfa00fba9c92ccf28d38606ab967ed19).

Switching to type aliasing can help us reduce duplication and should reduce growing pains as we stabilize more schemas. Once a type is defined, future schema's can then build a type alias from it, and if need be can then expand its functionality for that schema forward. This should then reduce the amount of redundant tests and struct definitions per schema. 

Looking at the fcos 1_4 schema.go, we have 4 structs, all of which are present in the previous schema.go 1_3. 

So as a simple example of what aliasing can do, instead of having: 
```
package v1_4

import (
	base "github.com/coreos/butane/base/v0_4"
)

type Config struct {
	base.Config `yaml:",inline"`
	BootDevice  BootDevice `yaml:"boot_device"`
}

type BootDevice struct {
	Layout *string          `yaml:"layout"`
	Luks   BootDeviceLuks   `yaml:"luks"`
	Mirror BootDeviceMirror `yaml:"mirror"`
}

type BootDeviceLuks struct {
	Tang      []base.Tang `yaml:"tang"`
	Threshold *int        `yaml:"threshold"`
	Tpm2      *bool       `yaml:"tpm2"`
}

type BootDeviceMirror struct {
	Devices []string `yaml:"devices"`
}
```

We could have: 

```
package v1_4

import (
	"github.com/coreos/butane/config/fcos/v1_3"
)

type Config v1_3.Config

type BootDevice v1_3.BootDevice

type BootDeviceLuks v1_3.BootDeviceLuks

type BootDeviceMirror v1_3.BootDeviceMirror

```

Showing a more complicated example would be the transition from 1_4 to 1_5_exp.

```
package v1_5_exp

import (
	base "github.com/coreos/butane/base/v0_5_exp"
)

type Config struct {
	base.Config `yaml:",inline"`
[...]

PR #372, opened 2022-07-15
WIP: Switch to type aliasing

prestist:
https://github.com/coreos/butane/issues/371


prestist:
Still very much a WIP, kinda racking my head on if type aliasing it really a good idea.

Issue #211, opened 2021-03-10
Warn if the root filesystem is too small


bgilbert:
Add a warning if the root filesystem is too small (https://github.com/coreos/fedora-coreos-tracker/issues/586), either because the underlying storage is explicitly specified, or because the root partition is constrained by an additional partition.

jlebon:
We've decided to defer this until we have https://github.com/coreos/fedora-coreos-tracker/issues/759 because otherwise it's impossible in many cases for Butane to know that it's looking at the root disk. There are some cases where it could (e.g. if root reprovisioning/mirroring is requested), but the common case of just adding a "trapping partition"  wouldn't be detectable without some dubious heuristics.

prestist:
This issue might belong in ignition? Similar to https://github.com/coreos/ignition/issues/1397#issuecomment-1154390522 
 With this change https://github.com/coreos/fedora-coreos-config/pull/1621

bgilbert:
Ignition is distro-independent and doesn't know the partition layout constraints of the OS it runs on, so this needs to go in Butane in the `fcos` variant.

prestist:
> Add a warning if the root filesystem is too small ([coreos/fedora-coreos-tracker#586](https://github.com/coreos/fedora-coreos-tracker/issues/586)), either because the underlying storage is explicitly specified, or because the root partition is constrained by an additional partition.

I wanted to ask a clarifying question, do we consider root constrained if there is a partition in front of it aka /var on number 5 ?

or is it only constrained if the filesystem has been specified and there is a partition in front of it ?

In speaking with @jlebon it is fairly common ask to have a /var partition after root for us.

bgilbert:
It's common to have a `/var` partition after the root partition, but we should only warn if that would make the root partition too small.  In sum, the values we need to check are:

- The `sizeMiB` of the root partition, if specified
- The `startMiB` of any following partition.  If this is unspecified (`0`) and the root `sizeMiB` is also unspecified, then the following partition will prevent the root partition from growing at all, and we should warn.

PR #378, opened 2022-08-04
fcos translate.go: add warn on small or constrained root partition

prestist:
https://github.com/coreos/butane/issues/211

Note I somehow dropped the '1' in the version for FCOS in the commit messages.. but before I update the commit structure lets make sure we are happy with the functional changes. Look at the first 2 commits, i.e the exp changes this is basically propagated to all versions back to 1_3 as you can see later with the fixup commits. Once approved I will merge the fixup commits with a rename to the commits to be inclusive of the changes!

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -68,6 +68,8 @@ var (
 	// partition
 	ErrReuseByLabel         = errors.New("partitions cannot be reused by label; number must be specified except on boot disk (/dev/disk/by-id/coreos-boot-disk) or when wipe_table is true")
 	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrRootTooSmall         = errors.New("root should have 8GiB of space assigned")
+	ErrRootConstrained      = errors.New("root partition cannot expand; it is set to fill available space but is followed by an auto-positioned partition")
 
 	// MachineConfigs
 	ErrFieldElided              = errors.New("field ignored in raw mode")
--- a/config/fcos/v1_7/translate.go
+++ b/config/fcos/v1_7/translate.go
@@ -71,16 +71,34 @@ func (c Config) ToIgn3_6Unvalidated(options common.TranslateOptions) (types.Conf
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
 	for i, disk := range ret.Storage.Disks {
-		// In the boot_device.mirror case, nothing specifies partition numbers
-		// so match existing partitions only when `wipeTable` is false
-		if !util.IsTrue(disk.WipeTable) {
-			for j, partition := range disk.Partitions {
-				// check for reserved partlabels
-				if partition.Label != nil {
-					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
-						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", j, "label"), common.ErrWrongPartitionNumber)
+		for p, partition := range disk.Partitions {
+			// check for root partition size constraints
+			if partition.Label != nil {
+				if *partition.Label == "root" {
+					if partition.SizeMiB == nil || *partition.SizeMiB == 0 {
+						for idx := range disk.Partitions {
+							if idx == p {
+								continue
+							}
+							if disk.Partitions[idx].StartMiB == nil || *disk.Partitions[idx].StartMiB == 0 {
+								r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "label"), common.ErrRootConstrained)
+								break
+							}
+						}
+					} else if *partition.SizeMiB < 8192 {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "size_mib"), common.ErrRootTooSmall)
 					}
 				}
+
+				// In the boot_device.mirror case, nothing specifies partition numbers
+				// so match existing partitions only when `wipeTable` is false
+				if !util.IsTrue(disk.WipeTable) {
+					// check for reserved partlabels
+					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "label"), common.ErrWrongPartitionNumber)
+					}
+				}
+
 			}
 		}
 	}
--- a/config/fcos/v1_3/translate.go
+++ b/config/fcos/v1_3/translate.go
@@ -70,16 +70,34 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 	}
 	r.Merge(c.processBootDevice(&ret, &ts, options))
 	for i, disk := range ret.Storage.Disks {
-		// In the boot_device.mirror case, nothing specifies partition numbers
-		// so match existing partitions only when `wipeTable` is false
-		if !util.IsTrue(disk.WipeTable) {
-			for j, partition := range disk.Partitions {
-				// check for reserved partlabels
-				if partition.Label != nil {
-					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
-						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", j, "label"), common.ErrWrongPartitionNumber)
+		for p, partition := range disk.Partitions {
+			// check for root partition size constraints
+			if partition.Label != nil {
+				if *partition.Label == "root" {
+					if partition.SizeMiB == nil || *partition.SizeMiB == 0 {
+						for idx := range disk.Partitions {
+							if idx == p {
+								continue
+							}
+							if disk.Partitions[idx].StartMiB == nil || *disk.Partitions[idx].StartMiB == 0 {
+								r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "label"), common.ErrRootConstrained)
+								break
+							}
+						}
+					} else if *partition.SizeMiB < 8192 {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "size_mib"), common.ErrRootTooSmall)
 					}
 				}
+
+				// In the boot_device.mirror case, nothing specifies partition numbers
+				// so match existing partitions only when `wipeTable` is false
+				if !util.IsTrue(disk.WipeTable) {
+					// check for reserved partlabels
+					if (*partition.Label == "BIOS-BOOT" && partition.Number != 1) || (*partition.Label == "PowerPC-PReP-boot" && partition.Number != 1) || (*partition.Label == "EFI-SYSTEM" && partition.Number != 2) || (*partition.Label == "boot" && partition.Number != 3) || (*partition.Label == "root" && partition.Number != 4) {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "label"), common.ErrWrongPartitionNumber)
+					}
+				}
+
 			}
 		}
 	}
--- a/config/fcos/v1_3/translate_test.go
+++ b/config/fcos/v1_3/translate_test.go
@@ -1423,3 +1423,242 @@ func TestTranslateBootDevice(t *testing.T) {
 		})
 	}
 }
+
+func TestRootPartitionConstraints(t *testing.T) {
+	tests := []struct {
+		name   string
+		in     Config
+		report report.Report
+	}{
+		{
+			name: "root constrained by auto-positioned partition",
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device: "/dev/vda",
+								Partitions: []base.Partition{
+									{
+										Label:   util.StrToPtr("root"),
+										Number:  4,
+										SizeMiB: util.IntToPtr(0), // fill available
+									},
+									{
+										Label:    util.StrToPtr("data"),
+										StartMiB: util.IntToPtr(0), // auto-positioned - will be placed after root
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			report: report.Report{
+				Entries: []report.Entry{
+					{
+						Kind:    report.Warn,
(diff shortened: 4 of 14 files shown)


config/fcos/v1_5_exp/translate.go
```diff
@@ -68,17 +68,31 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
...
-				// check for reserved partlabels
-				if partition.Label != nil {
+		for p, partition := range disk.Partitions {
+			if partition.Label != nil {
+				if *partition.Label == "root" {
+					if partition.SizeMiB != nil && *partition.SizeMiB < 8192 {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "sizeMiB", *partition.SizeMiB), common.ErrRootTooSmall)
+					} else {
+						for _, op := range disk.Partitions {
+							if op.Number > partition.Number {
```
bgilbert, line 80: We're currently performing this check for every partition after the root partition, not just the one immediately after it.

config/fcos/v1_5_exp/translate.go
```diff
@@ -68,17 +68,31 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
...
+								if op.StartMiB == nil {
```
bgilbert, line 81: This doesn't match the logic described in https://github.com/coreos/butane/issues/211#issuecomment-1205967355.

config/common/errors.go
```diff
@@ -37,6 +37,8 @@ var (
 	ErrNodeExists             = errors.New("matching filesystem node has existing contents or different type")
 	ErrNoFilesDir             = errors.New("local file paths are relative to a files directory that must be specified with -d/--files-dir")
 	ErrTreeNotDirectory       = errors.New("root of tree must be a directory")
+	ErrRootTooSmall           = errors.New("root should have 8GiB of space available")
+	ErrRootNotLastPartition   = errors.New("root should be last partition number to allow for growth")
```
bgilbert, line 41: The message is misleading, since we allow and encourage adding partitions after the root.

config/common/errors.go
```diff
@@ -37,6 +37,8 @@ var (
 	ErrNodeExists             = errors.New("matching filesystem node has existing contents or different type")
 	ErrNoFilesDir             = errors.New("local file paths are relative to a files directory that must be specified with -d/--files-dir")
 	ErrTreeNotDirectory       = errors.New("root of tree must be a directory")
+	ErrRootTooSmall           = errors.New("root should have 8GiB of space available")
```
bgilbert, line 40: nit: `available` could be read to imply free space in the filesystem.

config/fcos/v1_5_exp/translate.go
```diff
@@ -68,17 +68,31 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
...
 	for i, disk := range ret.Storage.Disks {
-		// In the boot_device.mirror case, nothing specifies partition numbers
-		// so match existing partitions only when `wipeTable` is false
-		if !util.IsTrue(disk.WipeTable) {
-			for j, partition := range disk.Partitions {
-				// check for reserved partlabels
-				if partition.Label != nil {
+		for p, partition := range disk.Partitions {
+			if partition.Label != nil {
+				if *partition.Label == "root" {
```
bgilbert, line 75: Unnecessary nesting.

config/fcos/v1_5_exp/translate.go
```diff
@@ -68,17 +68,31 @@ func (c Config) ToIgn3_4Unvalidated(options common.TranslateOptions) (types.Conf
...
+						for _, op := range disk.Partitions {
+							if op.Number > partition.Number {
+								if op.StartMiB == nil {
+									r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "number", partition.Number), common.ErrRootNotLastPartition)
+								}
+							}
+						}
+					}
+				}
+				// Don't warn if wipeTable is set, matching later spec versions
```
bgilbert, line 88: Looks like we lost the original comment here.  "matching later spec versions"?

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -146,6 +146,198 @@ func TestTranslateBootDevice(t *testing.T) {
...
+				{path.New("yaml", "storage", "filesystems", 0), path.New("json", "storage", "filesystems", 0)},
+				{path.New("yaml", "storage", "filesystems"), path.New("json", "storage", "filesystems")},
+				{path.New("yaml", "storage"), path.New("json", "storage")},
+			},
+			report.Report{
+				Entries: []report.Entry{
+					{
+						Kind:    report.Warn,
+						Message: common.ErrRootTooSmall.Error(),
+						Context: path.New("json", "storage", "disks", 0, "partitions", 0, "sizeMiB", 500),
```
bgilbert, line 239: This is specifying the 501st element of the `sizeMiB` array.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -146,6 +146,198 @@ func TestTranslateBootDevice(t *testing.T) {
...
+					Version: "3.4.0-experimental",
+				},
+				Storage: types.Storage{
+					Disks: []types.Disk{
+						{
+							Device: "/dev/vda",
+							Partitions: []types.Partition{
+								{
+									Label:   util.StrToPtr("root"),
+									SizeMiB: util.IntToPtr(10000),
```
bgilbert, line 290: This size is sufficient, so the test should pass.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -146,6 +146,198 @@ func TestTranslateBootDevice(t *testing.T) {
...
+			report.Report{
+				Entries: []report.Entry{
+					{
+						Kind:    report.Warn,
+						Message: common.ErrRootNotLastPartition.Error(),
+						Context: path.New("json", "storage", "disks", 0, "partitions", 0, "number", 4),
+					},
+				},
+			},
+		},
```
bgilbert, line 340: Let's also add tests where the root partition doesn't explicitly specify a size but is constrained by another partition which does/doesn't leave enough room.

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -146,6 +146,198 @@ func TestTranslateBootDevice(t *testing.T) {
...
+				{path.New("yaml", "storage", "filesystems", 0), path.New("json", "storage", "filesystems", 0)},
+				{path.New("yaml", "storage", "filesystems"), path.New("json", "storage", "filesystems")},
+				{path.New("yaml", "storage"), path.New("json", "storage")},
+			},
+			report.Report{
+				Entries: []report.Entry{
+					{
+						Kind:    report.Warn,
+						Message: common.ErrRootNotLastPartition.Error(),
+						Context: path.New("json", "storage", "disks", 0, "partitions", 0, "number", 4),
```
bgilbert, line 336: Likewise, this is the 5th element of the `number` array.

config/fcos/v1_5_exp/translate.go
prestist: OH can we assume that they are always in order?

bgilbert: What we don't want is to warn repeatedly for every additional partition.  But you're right, we can't assume they're in order, either by starting offset or by partition number.  And actually, we can't make assumptions about `op.Number` either, since it might be 0.

So we'll need an heuristic for figuring out the earliest partition start.  If the `startMiB` is always specified, this is easy.  Otherwise, the partition order is determined by a combination of `sgdisk`'s heuristics (see the man page for the `-n` option) and the additional heuristics applied by [partitions.go](https://github.com/coreos/ignition/blob/main/internal/exec/stages/disks/partitions.go).  And unfortunately, they have some information we don't, which is the overall size of the disk, and thus the size of the largest available block.

So we'll need to figure this out.  If there are cases where we're not sure, it's better to not warn.

aaradhak:
The first commit message says fcos v0_7_exp but there is no v0_7_exp directory. May be this should be fcos v1_7_exp?

aaradhak:
Minor: Typo in commit message body: "contstraints" should be "constraints"

config/common/errors.go
```diff
@@ -68,6 +68,8 @@ var (
 	// partition
 	ErrReuseByLabel         = errors.New("partitions cannot be reused by label; number must be specified except on boot disk (/dev/disk/by-id/coreos-boot-disk) or when wipe_table is true")
 	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrRootTooSmall         = errors.New("root should have 8GiB of space assigned")
+	ErrRootConstrained      = errors.New("root is configured too small, and has no room to expand")
```
aaradhak: just curious if this can this be rephrased? shouldnt this error message say something like that it's configured to fill available space?

config/fcos/v1_7/translate.go
```diff
@@ -71,16 +71,70 @@ func (c Config) ToIgn3_6Unvalidated(options common.TranslateOptions) (types.Conf
...
+							}
+						}
+					} else if *partition.SizeMiB < 8192 {
+						r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "size_mib"), common.ErrRootTooSmall)
+					}
+				}
+
+				// In the boot_device.mirror case, nothing specifies partition numbers
+				// so match existing partitions only when `wipeTable` is false
+				if !util.IsTrue(disk.WipeTable) {
```
aaradhak, line 131: Trying to understand this wipetable change. The previous code only checked reserved partlabels when wipeTable is false. The current change now runs the root validation even when wipeTable is true. I guess it is intentional.

prestist:
@aaradhak  

Fixed both issues! I've rebased and updated the commit messages:

  1. Changed from "fcos v0_7_exp" to "fcos translate.go" / "fcos translate_test.go" since
     these changes actually apply to FCOS versions v1_3 through v1_7

  2. Fixed the typo: "contstraints" → "constraints"

config/fcos/v1_7/translate.go
prestist: Yeah it was intentional the root partition size validation needs to happen regardless of wipeTable, because:

  1. Even when wipeTable is true (creating a fresh partition table), users can still specify
     partitions that trap the root partition - this is exactly the scenario we want to warn about

  2. The wipeTable check only applies to the "reserved partlabel validation" (checking that
     BIOS-BOOT is partition 1, root is partition 4, etc.). That validation should only run
     when wipeTable is false because we're matching existing partitions.

  3. The root size/constraint validation is orthogonal to whether we're wiping the table -
     it's about warning users before they create a configuration that won't have enough space
     for OS updates.

config/common/errors.go
prestist: Good point! The current message "root is configured too small, and has no room to expand"
  could be clearer.

aaradhak:
@prestist  I happened to go over the release details of butane [v0.27.0](https://github.com/coreos/butane/releases/tag/v0.27.0) . I bumped into this ` Add Fcos spec 1.8.0-experimental` , Should this validation also be added to `config/fcos/v1_8_exp/translate.go`? Was it excluded intentionally?

prestist:
> @prestist I happened to go over the release details of butane [v0.27.0](https://github.com/coreos/butane/releases/tag/v0.27.0) . I bumped into this ` Add Fcos spec 1.8.0-experimental` , Should this validation also be added to `config/fcos/v1_8_exp/translate.go`? Was it excluded intentionally?

Nope that was an oversite, due to the latest stabilization and the age of this pr.

prestist:
@aaradhak 

If you want to test it locally you can do this:


```md
1. Checkout and Build                                                                                                                                                  
                                                                                                                                                                         
  # Checkout the PR                                                                                                                                                      
  gh pr checkout 378                                                                                                                                                     
                                                                                                                                                                         
  # Build butane                                                                                                                                                         
  ./build                                                                                                                                                                                                               
                                                                                                                                                                         
  2. Manual Testing Scenarios
                                                                                                                                                                         
  The PR adds two warnings for root partition issues:                                                                                                                    
  
  ⚠️  Warning 1: Root Too Small (< 8 GiB)                                                                                                                                 
                                                                            
  cat > test-small.bu << 'EOF'                                                                                                                                           
  variant: fcos                                                                                                                                                          
  version: 1.7.0
  storage:                                                                                                                                                               
    disks:                                                                  
      - device: /dev/vda
        wipe_table: false                                                                                                                                                
        partitions:
          - number: 4                                                                                                                                                    
            label: root                                                     
            size_mib: 5000                                                                                                                                               
  EOF                                                                                                                                                                    
  
  ./bin/amd64/butane < test-small.bu                                                                                                                                     
  Result: warning: root should have 8GiB of space assigned ✅               
                                                                                                                                                                         
  ⚠️  Warning 2: Root Constrained by Auto-Positioned Partition
                                                                                                                                                                         
  cat > test-constrained.bu << 'EOF'                                        
  variant: fcos                                                                                                                                                          
  version: 1.7.0                                                            
  storage:
    disks:
      - device: /dev/vda
        partitions:                                                                                                                                                      
          - number: 4                                                                                                                                                    
            label: root                                                                                                                                                  
            size_mib: 0                                                                                                                                                  
          - number: 5                                                       
            label: data
            start_mib: 0
  EOF                                                                                                                                                                    
                                                                                                                                                                         
  ./bin/amd64/butane < test-constrained.bu                                                                                                                               
  Result: warning: root partition cannot expand; it is set to fill available space but is followed by an auto-positioned partition ✅                                    
                                                                                                                                                                         
  ✅ No Warning: 8 GiB or larger
                                                                                                                                                                         
  cat > test-ok.bu << 'EOF'                                                 
  variant: fcos                                                                                                                                                          
  version: 1.7.0                                                            
  storage:
    disks:                                                                                                                                                               
      - device: /dev/vda
        wipe_table: false                                                                                                                                                
        partitions:                                                         
          - number: 4                                                                                                                                                    
            label: root                                                     
            size_mib: 8192
  EOF

  ./bin/amd64/butane --strict < test-ok.bu  # No warnings  

```

config/fcos/v1_7/translate.go
```diff
@@ -71,16 +71,70 @@ func (c Config) ToIgn3_6Unvalidated(options common.TranslateOptions) (types.Conf
...
+							if apStart > rootStart || (apStart == 0 && rootStart >= 0) {
+								if nextPartition == nil {
+									nextPartition = &ap
+								} else {
+									var nextStart int
+									if nextPartition.StartMiB != nil {
+										nextStart = *nextPartition.StartMiB
+									}
+									// If this partition starts before the current "next", it's closer to root
+									if (apStart > 0 && apStart < nextStart) || (nextStart == 0 && apStart > 0) {
```
aaradhak: The condition here causes an explicitly positioned partition to replace an auto-positioned one as the "nearest" candidate but the auto-positioned partition is the one that actually constrains root right?

prestist: oooof, okay; yees you are right; Let me update the code to account for this.

prestist:
@aaradhak Good catch! You're right - the previous logic missed when both an auto-positioned and an explicitly positioned partition were present, the "find nearest"  would prefer the explicit partitions, effectively hiding the auto-positioned partition that actually constrains root.                                                                                          
                                                                                                                                                                         
The startMiB of any following partition. If this is unspecified (0) and the root sizeMiB is also unspecified, then the following partition will prevent the root partition from growing at all, and we should warn.

Instead of trying to determine which partition is physically "nearest" to root (which we can't reliably do without knowing disk size and sgdisk's placement heuristics), we now simply check: if root is auto-sized AND any other partition on the disk has startMiB unspecified, warn. I also added a test case for this scenario (auto-positioned partition with an explicit one also present).

aaradhak (approved):
LGTM

Issue #393, opened 2022-09-28
`butane` hangs when executed without any argument


gmeghnag:
When executed without any argument `butane` should return how to use it, as example:
```
Usage: butane [options] [input-file]
Options:
  -d, --files-dir string   allow embedding local files from this directory
  -h, --help               show usage and exit
  -o, --output string      write to output file instead of stdout
  -p, --pretty             output formatted json
  -r, --raw                never wrap in a MachineConfig; force Ignition output
  -s, --strict             fail on any warning
  -V, --version            print the version and exit
```

Instead, **when executed without argument it hangs indefinitely**.

bgilbert:
This is working as intended.  Butane reads from standard input by default, following the convention of Unix command-line programs that process an input stream.

cgwalters:
Agree it's a Unix convention, but there is some precedent for doing different things if stdin/stdout is a tty.  I'm a bit skeptical that anyone is going to hand-type in YAML on a tty, so we could consider erroring out in that case and printing help.

PR #394, opened 2022-09-28
Fix issue #393

gmeghnag:
Show usage and exit when butane is executed without any arguments.


bgilbert:
Thanks for the PR.  This would break Butane reading from standard input, which is a supported workflow that's encouraged by our documentation, so I'll close.

Issue #312, opened 2022-01-25
`openshift` experimental spec uses an Ignition spec not supported by the MCO


bgilbert:
The MCO doesn't yet support Ignition spec 3.3.0 (xref https://github.com/openshift/machine-config-operator/pull/2675).  If the MCO isn't updated by the time the next `openshift` spec is stabilized, we need to roll back the latter to Ignition spec 3.2.0.

This was previously done in #269 and #307.  Note that it's more convenient to perform the rollback _after_ stabilizing the spec, since that leaves the new experimental spec still using the newer Ignition spec version.

PR #395, opened 2022-10-05
Stabilize openshift 4.12.0; create openshift 4.13.0-experimental

bgilbert:
The MCO doesn't support Ignition spec 3.3.0 yet, so we also need to [roll back](https://github.com/coreos/butane/issues/312) `openshift` 4.12.0 to `fcos` 1.3.0 and `base` 0.3.


prestist (approved):
LGTM

PR #401, opened 2022-10-25
fcos/v1_5_exp: check return value when writing to hash

bgilbert:
Hash writes shouldn't fail, but Coverity doesn't know that.  Add a check and panic just to be sure.


prestist (approved):
LGTM

PR #404, opened 2022-11-08
Add new r4e variant

### LINKED ISSUE #396: RFE: new variant supporting RHEL for Edge
Creating this issue as a place-holder for discussing a variant specific to the RHEL for Edge use case.

We envision supporting Ignition on RHEL for Edge systems in the future for handling simple configuration tasks like writing files and enabling systemd units.  We don't want to (immediately) support the full complement of Ignition features (i.e. disks, RAID, filesystems, LUKS) so it was suggested a new variant be created in butane to denylist the features we don't want to support.

cc: @runcom

7flying:
This PR adds a new rhel-for-edge variant `r4e` targeting ignition 3.3.0 as well as an experimental variant targeting ignition 3.4.0.

Closes #396 

Checklist:
*Right now we won't support:*
- [x]  storage.{disks, filesystems, luks, raid}
- [x] kernel_arguments

### DIFF
--- a/config/r4e/v1_0/translate.go
+++ b/config/r4e/v1_0/translate.go
@@ -1,0 +1,79 @@
+// Copyright 2022 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.)
+
+package v1_0
+
+import (
+	"github.com/coreos/butane/config/common"
+	cutil "github.com/coreos/butane/config/util"
+	"github.com/coreos/butane/translate"
+
+	"github.com/coreos/ignition/v2/config/v3_3/types"
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
+)
+
+// ToIgn3_3Unvalidated translates the config to an Ignition config.  It also
+// returns the set of translations it did so paths in the resultant config
+// can be tracked back to their source in the source config.  No config
+// validation is performed on input or output.
+func (c Config) ToIgn3_3Unvalidated(options common.TranslateOptions) (types.Config, translate.TranslationSet, report.Report) {
+	ret, ts, r := c.Config.ToIgn3_3Unvalidated(options)
+	if r.IsFatal() {
+		return types.Config{}, translate.TranslationSet{}, r
+	}
+
+	checkForForbiddenFields(ret, &r)
+
+	return ret, ts, r
+}
+
+// Checks and adds the appropiate errors when unsupported fields on r4e are
+// provided
+func checkForForbiddenFields(t types.Config, r *report.Report) {
+	for i := range t.KernelArguments.ShouldExist {
+		r.AddOnError(path.New("path", "json", "kernel_arguments", "should_exist", i), common.ErrGeneralKernelArgumentSupport)
+	}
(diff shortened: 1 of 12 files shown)


7flying:
@bgilbert please take a look

config/r4e/v1_1_exp/translate_test.go
```diff
@@ -0,0 +1,209 @@
...
+	type InvalidEntry struct {
+		Kind report.EntryKind
+		Err  error
+		Path path.ContextPath
+	}
+	tests := []struct {
+		In      Config
+		Entries []InvalidEntry
+	}{
+		{
```
prestist, line 40: It might make sense to add some high level comments around each config for glance value. 

i.e. https://github.com/coreos/butane/blob/main/config/fcos/v1_4/translate_test.go#L56

prestist (commented):
Generally looks good, just have a few questions.

config/r4e/v1_1_exp/translate_test.go
7flying: There is a comment on the test function explaining what the test is about: `// Test that we error on unsupported fields for r4e` 
Should I repeat something similar to that on each config?

7flying: I see! I agree that it would make it easier to jump into the code and understand the test. I will update it with comments :+1:

prestist (approved):
LGTM

config/r4e/v1_0/translate.go
```diff
@@ -0,0 +1,85 @@
...
+// Checks and adds the appropiate errors when unsupported fields on r4e are
+// provided
+func checkForForbiddenFields(t types.Config, r *report.Report) {
+	for i := range t.KernelArguments.ShouldExist {
+		r.AddOnError(path.New("path", "json", "kernel_arguments", "should_exist", i), common.ErrGeneralKernelArgumentSupport)
+	}
+	for i := range t.KernelArguments.ShouldNotExist {
+		r.AddOnError(path.New("path", "json", "kernel_arguments", "should_not_exist", i), common.ErrGeneralKernelArgumentSupport)
+	}
+	for i := range t.Storage.Directories {
```
bgilbert, line 51: It seems surprising that files are supported but directories and links aren't, since they're all in the class of "filesystem objects".  For example, Ignition will implicitly create parent directories for files, but without explicit `Directories` support there won't be a way of specifying their ownership or permissions.  What's the reasoning here?

docs/config-r4e-v1_0.md
```diff
@@ -0,0 +1,119 @@
+---
+title: RHEL for Edge v1.0.0
+parent: Configuration specifications
+nav_order: 149
```
bgilbert, line 4: 100-150 are already used by the OpenShift spec (`rhcos` is grouped with it for historical reasons).  Alphabetically, R4E should use 199 and 200.

docs/config-rhcos-v0_1.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: RHEL CoreOS v0.1.0
 parent: Configuration specifications
-nav_order: 149
+nav_order: 151
```
bgilbert, line 4: This should stay at 149.

7flying:
@prestist do you want me to update `docs/release-notes.md` or is it something that you folks do as part of the new release checklist?

prestist:
> LGTM

Whoops I lied, you need to add a comment for the Release notes, that describes the functionality added found [here](https://github.com/coreos/butane/blob/main/docs/release-notes.md).

bgilbert (commented):
Code LGTM on a quick skim.  Nice work!

config/r4e/v1_0/translate.go
7flying: let me ping @runcom and @miabbott so that we can see what we do here since that's a good point.

7flying: yep, we want those, my mistake, will update in a moment :+1:

prestist (approved):
LGTM! Thank you for being so quick to respond and make the appropriate changes!

PR #407, opened 2022-11-17
dockerfile: rebase to f37 from f36

prestist:
https://github.com/coreos/fedora-coreos-tracker/issues/1223


dustymabe (approved):
LGTM

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:36 AS builder
+FROM registry.fedoraproject.org/fedora:37 AS builder
```
jlebon, line 1: This needs to update the second `FROM` below too.

prestist: Thank you for catching that, I am making a pr now.

Issue #257, opened 2021-06-24
Remove `rhcos` 0.1.0 spec


bgilbert:
It was a preview implementation of an OpenShift-specific spec (thus 0.x) and has been superseded by the `openshift` variant.  It's used in the OCP 4.7 docs, so we should wait until 4.7 goes EOL, but we can probably remove it thereafter.

xref #256

PR #413, opened 2023-01-04
Remove deprecated `rhcos` variant

bgilbert:
It only had one version, 0.1.0, which was a preview implementation of an OpenShift-specific spec.  It was only used in OpenShift 4.7, which is now EOL, and Butane has emitted a deprecation warning upon use for a year.

This is an API break.

Fixes https://github.com/coreos/butane/issues/257.


cgwalters (approved):
:fire:

PR #415, opened 2023-01-10
Improve error on unsupported `rhcos` variant

bgilbert:
Now that the rhcos variant has been removed, attempting to use an rhcos config produces a non-actionable error:

    Error translating config: No translator exists for variant rhcos with version 0.1.0

Add an explicit check and message:

    Error translating config: rhcos variant has been removed; use openshift variant instead: https://coreos.github.io/butane/upgrading-openshift/


bgilbert:
No release note since the previous release note already covers this.

prestist (approved):
Lgtm

Issue #119, opened 2020-07-16
Cannot transpile config with storage tree of current directory


bostrt:
I'm sure there's another way I can organize my code, but I hit a snag while trying to using `./` as a storage tree value. I'm *honestly* not sure if this a use-case that fcct intends to support, but figured I would still an issue.

Basically, I have my `fcct-test.yaml`, `Dockerfile`, and some "blog" (example) code in same root folder. 

The reason I've filed this issue is that *technically*, `-d ./` suggests I can include local files from the current directory but its not being accepted. 

```
$ fcct --version
Fedora CoreOS Config Transpiler 0.6.0

$ ls 
blog-posts/  Dockerfile  fcct-test.yaml
$ cat fcct-test.yaml | fcct -d ./ --pretty --strict
error at $.storage.trees.0, line 26 col 7: local file path traverses outside the files directory
Error translating config: source config is invalid
```

```
$ cat fcct-test.yaml 
variant: fcos
version: 1.1.0
systemd:
  units:
    - name: build-blog.service
      enabled: true
      contents: |
        [Unit]
        Description=Build blog container
        After=network-online.target
        Wants=network-online.target
        [Service]
        # Run as non-root user
        User=core
        # Give time for the build to complete
        TimeoutStartSec=10m
        Type=oneshot
        RemainAfterExit=yes
        ExecStartPre=-podman pull docker.io/library/nginx
        ExecStart=-podman build -t localhost/blog:latest /home/core/blog-build/
        ExecStartPost=-podman image prune --force
        [Install]
        WantedBy=multi-user.target
storage:
  trees:
    - local: ./
      path: /home/core/blog-build/
```

bgilbert:
Yup, this is a bug.  It should be legal for `storage.trees.local` to be exactly equal to `FilesDir`, but `EnsurePathWithinFilesDir()` requires it to be a child of `FilesDir` instead.  Thanks for reporting.

PR #416, opened 2023-01-11
base/util: avoid error if storage.trees.local value equals to FilesDir

sohankunkerkar:
Fixes https://github.com/coreos/butane/issues/119

### DIFF
--- a/base/util/file.go
+++ b/base/util/file.go
@@ -30,7 +30,7 @@ func EnsurePathWithinFilesDir(path, filesDir string) error {
 	if err != nil {
 		return err
 	}
-	if !strings.HasPrefix(absPath, absBase+string(filepath.Separator)) {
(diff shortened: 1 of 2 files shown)


base/util/file.go
```diff
@@ -30,7 +30,7 @@ func EnsurePathWithinFilesDir(path, filesDir string) error {
 	if err != nil {
 		return err
 	}
-	if !strings.HasPrefix(absPath, absBase+string(filepath.Separator)) {
+	if strings.Compare(absPath, absBase) != 0 && !strings.HasPrefix(absPath, absBase+string(filepath.Separator)) {
```
bgilbert, line 33: Logic looks okay, but is there a reason to use `strings.Compare()` instead of `!=`?

sohankunkerkar: There's no particular reason as such. I usually prefer to use a built-in function.

bgilbert (approved):

PR #421, opened 2023-01-29
Support LUKS open options; add `boot_device.luks.discard` field

bgilbert:
Add LUKS `discard` and `open_options` fields from https://github.com/coreos/ignition/pull/1540.  Also, users might reasonably want to enable discard for rootfs LUKS, so add it to the `boot_device` sugar.  Don't add an `open_options` field there, since the sugar isn't meant to provide that level of customization.


bgilbert:
Okay, this should be ready!

prestist (approved):
Glad to see this; quite relevant! 

LGTM!!!

PR #424, opened 2023-02-05
base: use default dependencies for mount units

bgilbert:
In mount units for local filesystems, we've been enabling default dependencies but redundantly specifying `Before=local-fs.target`. Remove it.

In mount units for filesystems on Tang-unlocked LUKS devices, we've been disabling default dependencies and then manually specifying `Before=remote-fs.target`.  This omits sequencing with `umount.target` for umount at shutdown; ordering dependencies on `remote-fs-pre.target`, `network.target`, and `network-online.target`; and `Wants=remote-fs.target`. There was some discussion of the proper directives when LUKS mount unit support was added in https://github.com/coreos/butane/pull/112, but nothing to indicate that default dependencies won't work.  Set `Options=_netdev` and use default dependencies.


bgilbert:
@jlebon @dustymabe Any concerns?

jlebon:
Using the `_netdev` option is definitely cleaner. At the time this functionality was added, I was looking at the `systemd-fstab-generator` as reference and tried to match that since the use case was the rootfs and we rely on that generator there. But of course for bona fide mount units it makes sense to leverage the built-in semantics.

Were you able to test this on e.g. a `/var/somedir` Tang-pinned mount? I think we have a CI gap around this right now.

jlebon (approved):
Patch itself looks good!

bgilbert:
Yup, it works with Tang-pinned secondary filesystems. :+1:

jlebon (approved):

dustymabe:
> @jlebon @dustymabe Any concerns?

I trust you :)

Issue #314, opened 2022-01-25
openshift/v4_13_exp: revise extensions feature before stabilizing the API


jmarrero:
The extensions API added on: https://github.com/coreos/butane/pull/304 which is part of https://github.com/coreos/butane/issues/300 needs to be revisited and discussed before stabilizing the 1.5 API. This feature might conflict with the MCO extensions capabilities and needs to be further examined before it's considered stable.

Further more @bgilbert has raised https://github.com/coreos/butane/issues/312 which might block this from being used on Openshift.
More context on: https://github.com/coreos/butane/pull/304#discussion_r790336682 & https://github.com/coreos/butane/pull/304#discussion_r791004773

bgilbert:
I think it's okay to stabilize `fcos` 1.5.0 without this; the concern is stabilizing it into an `openshift` spec.

bgilbert:
We're getting close to stabilizing a 1.5.0 spec.  The functionality wouldn't be available in `openshift` 4.13.0 but would be available in 4.14.0.  @jmarrero, @jlebon, how do you feel about the state of this feature?

jmarrero:
I am torn... while the declarative interface of butane makes sense in our current world. I am not sure that with the discussions we are having about layering we want to expose this functionality to openshift just yet. Might be worth adding it to our discussions on layering/bootc/ignition future.

travier:
Agree that we should remove that from the about to be stabilized spec and defer it for the next one.

bgilbert:
@jlebon, @jmarrero, @travier, and I discussed this OOB and agreed that we should back out the `extensions` sugar for now.  It seems like a pretty narrow UX, and may not be a good fit for the layering use cases generally.  @cgwalters, any objection?

cgwalters:
Makes sense to me.

PR #425, opened 2023-02-08
fcos/v1_5_exp: drop extensions section

bgilbert:
It's not clear that we want this as-is, so avoid stabilizing it for now.

Removal of `ErrExtensionNameRequired` is technically a breaking change, but it seems unlikely that any caller is checking for that error.

Fixes https://github.com/coreos/butane/issues/314.


jmarrero (approved):
lgtm

prestist (approved):

travier (approved):
Approved in https://github.com/coreos/butane/issues/314#issuecomment-1424318313

PR #429, opened 2023-03-02
Tang offline provisoning

prestist:


config/fcos/v1_5_exp/translate_test.go
```diff
@@ -219,6 +219,73 @@ func TestTranslateBootDevice(t *testing.T) {
...
+							WipeFilesystem: util.BoolToPtr(true),
+						},
+					},
+				},
+			},
+			[]translate.Translation{
+				{From: path.New("yaml", "version"), To: path.New("json", "ignition", "version")},
+				{From: path.New("yaml", "boot_device", "luks", "tang", 0, "url"), To: path.New("json", "storage", "luks", 0, "clevis", "tang", 0, "url")},
+				{From: path.New("yaml", "boot_device", "luks", "tang", 0, "thumbprint"), To: path.New("json", "storage", "luks", 0, "clevis", "tang", 0, "thumbprint")},
+				{From: path.New("yaml", "boot_device", "luks", "tang", 0, ""), To: path.New("json", "storage", "luks", 0, "clevis", "tang", 0, "advertisement")},
```
prestist, line 269: I am surprised by this, I expected `{From:....., "tang",0,"advertisement"} `but instead I got `{From:....., "tang",0,""} `
Is that an indication of something wrong? @bgilbert

bgilbert:
Nits for commit messages:

1. The commit hash isn't super-important (it's in `go.mod`) so let's drop it to avoid an over-length commit title.  (And also clean up the wording a bit.)
2. Prefix with `base/v0_5_exp:`
3. Drop `openshift/v_13_exp`.  The OpenShift spec is just inheriting this without code changes.  (Also, would be `v4_13_exp`.)  And we're not really adding the field, since it's inherited from base; we're just adding tests.

base/v0_5_exp/schema.go
```diff
@@ -227,8 +227,9 @@ type Systemd struct {
 }
 
 type Tang struct {
-	Thumbprint *string `yaml:"thumbprint"`
-	URL        string  `yaml:"url"`
+	Thumbprint    *string `yaml:"thumbprint"`
+	URL           string  `yaml:"url"`
+	Advertisement *string `json:"advertisement,omitempty"`
```
bgilbert, line 232: `omitempty` isn't necessary for pointers.

docs/release-notes.md
```diff
@@ -16,6 +16,8 @@ nav_order: 9
   flatcar 1.1.0-exp, openshift 4.13.0-exp)_
 - Allow specifying arbitrary LUKS open options _(fcos 1.5.0-exp,
   flatcar 1.1.0-exp, openshift 4.13.0-exp)_
+- Allow enabling Tang offline provisioning _(fcos 1.5.0-exp,
+  flatcar 1.1.0-exp, openshift 4.13.0-exp)_
```
bgilbert, line 20: `flatcar` doesn't support Clevis, so that part should be dropped.

config/fcos/v1_5_exp/translate_test.go
bgilbert: Yes, that shouldn't be happening.  It's because you're tagging the `Advertisement` field in `schema.go` with `json:` instead of `yaml:`.

bgilbert (commented):
Actually, let's just combine the latter two commits.  Adding the field to `base` automatically adds it to the distro-specific `config` schemas, so I don't think there's a clear logical separation between the two steps.

docs/release-notes.md
```diff
@@ -16,6 +16,7 @@ nav_order: 9
   flatcar 1.1.0-exp, openshift 4.13.0-exp)_
 - Allow specifying arbitrary LUKS open options _(fcos 1.5.0-exp,
   flatcar 1.1.0-exp, openshift 4.13.0-exp)_
+- Allow enabling Tang offline provisioning _(fcos 1.5.0-exp, openshift 4.13.0-exp)_
```
bgilbert, line 19: The upstream release note was `Support offline Tang provisioning via pre-shared advertisement`.  We could reuse that here.

bgilbert (approved):
Looks good!

docs/release-notes.md
prestist: Ah, yeah thats fair. Wasn't sure if we were being purposeful about 'Allow'. though might as well keep the old one.

bgilbert: Not especially purposeful about it, no.

bgilbert (approved):

PR #431, opened 2023-03-02
openshift/v4_13_exp: allow `passwd.users.password_hash`

bgilbert:
The MCO added support in https://github.com/openshift/machine-config-operator/pull/3539.

Awkwardly add `"password_hash" (4.13.0+)` to the existing error string. The alternative is creating a new error code for 4.13+, which would produce a cleaner error message, but that would break any callers programmatically checking for `ErrUserFieldSupport`.

cc @jkyros


jkyros:
Thank you for doing this on such short notice.  I am sorry about the awkwardness here. :cry:

We'll do a better job of keeping y'all in the loop so we can engineer these things more holistically. 
 
/lgtm

prestist (approved):
LGTM!

PR #432, opened 2023-03-03
Reference to online converter.

meiXXI:


bgilbert:
I think we probably wouldn't want to encourage use of a web-based Butane converter hosted outside the Fedora project.  Butane configs often contain credentials or other sensitive information, so there's a risk that the service could be compromised and used to steal sensitive data.  This isn't a comment about you &mdash; no doubt your intentions are good &mdash; but we also need to consider security vulnerabilities, including in hosting providers beyond your control.

(One example of such an issue: the converter currently writes the user's Butane config to a temporary file, and then [fails to remove](https://github.com/meiXXI/butane-converter/blob/edb2989cc8ae03908182706937de84a821f54aca/api/app.go#L99-L103) that file if Butane produces any errors or warnings.)

In addition, Butane supports inclusion of local files via the `local` directive and `storage.trees` section, coupled with the `--files-dir` argument.  Online converters can't use these, of course, which reduces their usefulness for complex configs.

I'll go ahead and close this, but thanks for your interest in making Butane easier to use.

meiXXI:
Hi @bgilbert, many thanks for your feedback!

PR #435, opened 2023-03-06
Stabilize Ignition spec 3.4.0

prestist:
`go mod get ...ignition/v2@main`
`go mod tidy`
`go vendor`

update vendor'd ignition package to prepare for stabilizing openshift and fcos configs.


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/clarketm/json v1.17.1
 	github.com/coreos/go-semver v0.3.1
 	github.com/coreos/go-systemd/v22 v22.5.0
-	github.com/coreos/ignition/v2 v2.14.1-0.20230218211505-da6c29bdd24f
+	github.com/coreos/ignition/v2 v2.15.1-0.20230306202950-5c31ef8eb367
```
bgilbert, line 9: Hmm, why revendor to Git `main` rather than the latest release?  If we pin to a Git commit, Dependabot won't automatically update to new releases in the future.

base/v0_5_exp/translate.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/coreos/go-systemd/v22/unit"
 	"github.com/coreos/ignition/v2/config/util"
-	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
```
bgilbert, line 31: For legibility reasons, revendoring is an exception to the rule that every commit should be an independent change.  Let's split out the revendoring, even though the resulting commit won't build.

bgilbert:
I concur that this doesn't need release notes.  We'll note the changes during the Butane spec stabilizations.

bgilbert (approved):

Issue #434, opened 2023-03-06
Stabilize fcos 1.5.0 


prestist:
Pending https://github.com/coreos/butane/pull/289

Following the steps found in https://coreos.github.io/butane/development/ re stabilization.

travier:
yes, that's the idea

travier:
https://coreos.github.io/butane/specs/#butane-specifications-and-ignition-specifications

travier:
CC @runcom for R4E

7flying:
So in edge we have all the docs already asking the users to use the `rfe` variant at version 1.0.0 which targets Ignition 3.3.0.
We can stabilise 1.1.0 since the subset of options that we support is the same one as 1.0.0 has.

travier:
RHEL 9.2 will have Ignition 2.15 that will be able to use the 3.4.0 Ignition spec. Not sure how much of an improvement that is for R4E.

7flying:
> RHEL 9.2 will have Ignition 2.15 that will be able to use the 3.4.0 Ignition spec. Not sure how much of an improvement that is for R4E.

Is that Ignition 2.15 still able to handle Ignition spec 3.3.0 if we say `r4e` v1.0.0? If not I will have to notify the docs team

bgilbert:
Yes.  Stabilizing a new spec does not affect existing specs.

bgilbert:
Thanks for the feedback @tormath1 and @7flying!  We'll plan to stabilize both the `flatcar` and `r4e` specs.

Issue #377, opened 2022-08-04
Warn if config attempts to reuse partition by label


bgilbert:
Consider this config:

```yaml
variant: fcos
version: 1.4.0
storage:
  disks:
    - device: /dev/sdb
      wipe_table: false
      partitions:
        - label: foo
```

When the system is first provisioned with a blank `/dev/sdb`, Ignition will create a partition `foo` starting 1 MiB into the disk and extending to the end of the disk.  If the system is then reprovisioned, Ignition will create a second partition `foo` in the first MiB of the disk:

```
Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048        33554398   16.0 GiB    8300  foo
   2              34            2047   1007.0 KiB  8300  foo
```

If the system is provisioned a third time, Ignition will fail.

This config appears to reuse the partition when reprovisioning, but actually doesn't, because Ignition matches existing partitions only by partition number and not by label (https://github.com/coreos/ignition/issues/1219).  We should fix that, but as a start, we can warn about it.

We should warn if a disk has `wipe_table: false` and any partitions have `number: 0`.  However, we need to special-case the root disk: since its partition table is rewritten on every reprovision, it's valid not to specify a partition number there, and we shouldn't warn.  Unfortunately, the only way to distinguish the root disk is if it's specified via the `/dev/disk/by-id/coreos-boot-disk` symlink, and legacy configs won't do that.

Proposal: add a warning in the experimental spec only, and only for devices other than `coreos-boot-disk`.  Phrase the message to make it clear that the warning can be avoided for the root disk by specifying it as `/dev/disk/by-id/coreos-boot-disk`.

PR #436, opened 2023-03-09
fcos/v1_5_exp: Identify and match existing partitions based on label and number

Adam0Brien:


config/fcos/v1_5_exp/translate.go
```diff
@@ -362,3 +362,28 @@ func buildGrubConfig(gb Grub) string {
...
+
+func arePartitionEqual(p1, p2 types.Partition) bool {
+    return p1.Label == p2.Label && p1.Number == p2.Number
+}
+
+
+func checkPartitionTable(disk string, wipeTable bool, partitions []types.Partition) error {
+   // if wipetable is true do nothing
+	if wipeTable {
+        return nil
```
bgilbert, line 374: The indentation here is failing CI tests.  You can run them locally with `./test`.  `gofmt -w <file>` will fix the formatting.

config/fcos/v1_5_exp/translate.go
```diff
@@ -362,3 +362,28 @@ func buildGrubConfig(gb Grub) string {
...
+    }
+   // check partitions labels and numbers and see if there are duplicates
+
+   // problably a faster way to do this... 
+    for i, p1 := range partitions {
+        for j := i + 1; j < len(partitions); j++ {
+            p2 := partitions[j]
+            if arePartitionEqual(p1, p2) {
+                return fmt.Errorf("Duplicate partition with label and number on disk. This may cause issues during reprovisioning. Specify the root disk as /dev/disk/by-id/coreos-boot-disk to avoid this warning.", p1.Label, p1.Number, disk)
```
bgilbert, line 383: All validation errors need to be defined in `errors.go` and referenced by name, so that users of the Go API can programmatically check for the error without hardcoding a string comparison.

Note that the code isn't actually checking for `/dev/disk/by-id/coreos-boot-disk` anywhere.

`Specify the root disk as ...` is technically correct, but the phrasing could mislead the user into thinking that they should specify _this_ disk as `coreos-boot-disk`, when in fact we don't know whether this disk is the boot disk or not.  Let's try to refine the phrasing a bit.

config/fcos/v1_5_exp/translate.go
```diff
@@ -362,3 +362,28 @@ func buildGrubConfig(gb Grub) string {
 	superUserCmd := fmt.Sprintf("set superusers=\"%s\"\n", strings.Join(allUsers, " "))
 	return "# Generated by Butane\n\n" + superUserCmd + strings.Join(cmds, "\n") + "\n"
 }
+
+func arePartitionEqual(p1, p2 types.Partition) bool {
+    return p1.Label == p2.Label && p1.Number == p2.Number
+}
+
+
+func checkPartitionTable(disk string, wipeTable bool, partitions []types.Partition) error {
```
bgilbert, line 371: This function isn't actually called anywhere?

Validation checks should be done in the validation pass if at all possible.  We have an unusual case here though: this is a validation that runs solely on structs defined in `base`, but the validation needs to be in `config/fcos`.  (Non-FCOS-derived distros don't have `/dev/disk/by-id/coreos-boot-disk` and so configs for those distros don't have a way to denote the boot disk.)  `config/fcos` can't define a `Validate()` method on `Disk` because that struct is defined in `base`, so we'll need to define `Validate()` on the top-level `Config` struct in `config/fcos` instead.

config/fcos/v1_5_exp/translate.go
```diff
@@ -362,3 +362,28 @@ func buildGrubConfig(gb Grub) string {
...
+	if wipeTable {
+        return nil
+    }
+   // check partitions labels and numbers and see if there are duplicates
+
+   // problably a faster way to do this... 
+    for i, p1 := range partitions {
+        for j := i + 1; j < len(partitions); j++ {
+            p2 := partitions[j]
+            if arePartitionEqual(p1, p2) {
```
bgilbert, line 382: This doesn't match the logic described in #377, and in fact won't warn on the example given there.  How did you arrive at this approach?

bgilbert (commented):
Thanks for working on this!

bgilbert:
FYI, #434 will land first, so you'll eventually have to rebase this onto `fcos/v1_6_exp`.

config/fcos/v1_5_exp/translate.go
Adam0Brien: Oh I thought the labels and numbers of each partition could not be the same, ill change it so it checks if the number of the partition table is 0 and wipe_tables is false and warn if that's the case

config/fcos/v1_5_exp/translate.go
Adam0Brien: I have the new error message committed in config/common/errors.go now!

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1632,3 +1632,28 @@ func TestTranslateGrub(t *testing.T) {
 		})
 	}
 }
+
```
Adam0Brien, line 1635: Not 100% confident on this test. Probably needs a second look if anyone has the time :)

config/fcos/v1_5_exp/translate_test.go
```diff
@@ -1632,3 +1632,28 @@ func TestTranslateGrub(t *testing.T) {
...
+				Label:  util.StrToPtr("baz"),
+				Number: 2,
+			},
+		},
+	}
+
+	if err := isPartitionValid(disk); err != common.ErrInvalidPartition {
+		t.Errorf("Expected ErrInvalidPartition, but got %v", err)
+	}
+}
```
Adam0Brien, line 1659: When running this test for some reason it will only pass if a partition or partitions have the number 0 assigned to it..
will take a look and see where i went wrong

config/fcos/v1_5_exp/translate.go
bgilbert: There are already checks for overlapping labels/numbers, but confusingly they're [in a different repo](https://github.com/coreos/ignition/blob/5c31ef8eb3670e24bc4a02dc465d82107c2913a1/config/v3_5_experimental/types/disk.go#L29-L49).  (Ignition has its own config validation, and Butane runs it after translating the config.)

Adam0Brien:
Test is working fine now : fails when a partition has the number 0 and wipe_tables is set to false

bgilbert:
Have you manually tested with the config from #377?

Adam0Brien:
I haven't only the test that is in translate_test.go and i'm not entirely sure how i would go about testing manually

bgilbert:
You should be able to just run Butane and pass it the config, and see whether it outputs the warning.

Adam0Brien:
this is the output i got :

adamobrien@fedora:~/coreos/coreos-butane/butane(p-tables-tests⚡) » butane config.yaml                                                                                                                                                    2 ↵
{"ignition":{"version":"3.3.0"},"storage":{"disks":[{"device":"/dev/sdb","partitions":[{"label":"foo"}],"wipeTable":false}]}}

looks okay

bgilbert:
There's no warning there.

Adam0Brien:
There's not supposed to be a warning is there?

Adam0Brien:
my bad i thought there shouldn't be one taking a look now!

Adam0Brien:
okay so after running this config on a local build 

`variant: fcos
version: 1.4.0
storage:
  disks:
    - device: /dev/sdb
      wipe_table: false
      partitions:
        - label: foo
        - number: 1
`
generates a .ign fine

but if i was to change the number from 1 -> 0 then i get the following error
`error at $.storage.disks.0.partitions.1, line 9 col 11: a partition number >= 1 or a label must be specified
Error translating config: config generated was invalid
`

bgilbert:
You're declaring two separate partitions but I think you're intending to declare one.  You'll probably want to drop the `-` from the last line.

Adam0Brien:
my bad didn't know that made a new partition. tested it without the `-` and i didn't get an error...

config.yaml
```diff
@@ -0,0 +1,9 @@
+variant: fcos
+version: 1.5.0-experimental
+storage:
+  disks:
+    - device: /dev/sdb
+      wipe_table: false
+      partitions:
+        - label: foo
+          number: 0
```
Adam0Brien, line 9: accidental commit but this is the config that gave the error

config/fcos/v1_5_exp/translate.go
```diff
@@ -16,6 +16,7 @@ package v1_5_exp
 
 import (
 	"fmt"
+	"log"
```
Adam0Brien, line 19: probably a better way to do this without importing log will take a look for a better solution

Adam0Brien:
At the moment if i use this config

```variant: fcos
version: 1.5.0-experimental
storage:
  disks:
    - device: /dev/sdb
      wipe_table: false
      partitions:
        - label: foo```
         
butane will return the error message "Partition number cannot be zero.% "

works as intended but id love some feedback

File: config/fcos/v1_5_exp/schema.go [hunk 9]
```diff
@@ -49,3 +49,20 @@ type GrubUser struct {
 	Name         string  `yaml:"name"`
 	PasswordHash *string `yaml:"password_hash"`
 }
+
+type Disk struct {
```
prestist, line 53: Why are we adding Disk, Partiton to the fcos config? 

We have access to those types from the 
```
type Config struct {
	base.Config `yaml:",inline"`
	Grub        Grub       `yaml:"grub"`
}
```

config/fcos/v1_5_exp/validate_test.go
```diff
@@ -231,3 +231,40 @@ func TestValidateGrubUser(t *testing.T) {
...
 	}
 }
+
+func TestValidateDisk(t *testing.T) {
+	tests := []struct {
+		in      Disk
+		out     error
+		errPath path.ContextPath
+	}{
+		// valid disk
```
prestist, line 241: We probably want some negative testing as well, try and think of some different permutations that cause different results that goes through this newly added code.

config/fcos/v1_5_exp/validate.go
```diff
@@ -51,3 +51,14 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (disk Disk) Validate(c path.ContextPath) (r report.Report) {
```
prestist, line 55: While this logic seems to be checking for a partition number == 0 and wipe-table == false
What are we going to do for the special case of coreos-boot-disk?

config/common/errors.go
```diff
@@ -51,7 +51,8 @@ var (
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
 	// partition
-	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrWrongPartitionNumber   = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrInvalidPartitionNumber = errors.New("partition number cannot be zero while wipe_table is set to false")
```
prestist, line 55: From the reading of the issue, we need to word this in a way that "...its clear that the warning can be avoided for the root disk by specifying it as /dev/disk/by-id/coreos-boot-disk."

config/fcos/v1_5_exp/schema.go
Adam0Brien: I added it because when i tried to call Disk or Partition in validate.go or validate_test.go it would give me a "undeclared name" error on vscode. When i added the structs to schema.go it started working.
Is there a way to call the structs without adding them to the schema.go?

config/common/errors.go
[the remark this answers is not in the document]
Adam0Brien: Makes sense ill push a change!

config/fcos/v1_5_exp/schema.go
bgilbert: https://github.com/coreos/butane/pull/436#discussion_r1131501088

config/fcos/v1_5_exp/validate_test.go
```diff
@@ -231,3 +231,102 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			err:     common.ErrInvalidPartitionNumber,
+			errPath: path.New("yaml"),
+		},
+	}
+
```
Adam0Brien, line 323: this seems to cause the test to fail not sure where i'm going wrong here

config/fcos/v1_5_exp/validate.go
Adam0Brien: have that added now, anytime the root disk is specified it skips the rest of the test

config/fcos/v1_5_exp/validate_test.go
Adam0Brien: added a negative part to the test but it seems to make the test fail altogether, not sure were i'm going wrong will look into it!

config/fcos/v1_5_exp/validate.go
prestist: Take a look at the `Config` object and see https://github.com/coreos/butane/pull/436#discussion_r1131501088

prestist:
Hey @Adam0Brien you have been making a lot of headway! Keep it up! 

I was thinking it might make sense to clean up the commit history a bit, firstly some pretext:
This workflow is one way of many to contribute to projects. There are other workflows but we find this workflow works for us. (Man that was a lot of "work" to write that sentence ;P ) 

Our MR workflow heavily incorporates force push [(git push -f)](https://git-scm.com/docs/git-push). Which really means we are being a little bit more intentional about the commits that is pushed to our MRs. This is because our MR's dont [squash](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/squashing-commits). So all our commits that are in the MR will be present in the main branch. A consequence of that means we need to keep our working MR's quite tidy and our commits intentional. 

An example of common itteration of work would look like :

1. Do work
- git checkout -b "foo-test"
- git add config/new-work.txt
- git commit -m "new-work: add new-work file"
- git add config/changes.txt
- git commit -m "config/changes: changed some things"
- git push
2. Get feedback
- Make changes
- git log (to get hash of commit you want to fixup)
- git add config/changes.txt
- git commit --fixup <git hash of commit found above>
- git rebase -i --autosquash <git hash of commit found above>
- git push -f
3. iterate

Generally commits should:
- compile
- have concise messages
- comments should not exceed 80 characters per line
- explain the the work in present tense.  
- if fixing a bug or issue; ref it with [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

bgilbert:
I really like [git-interactive-rebase-tool](https://github.com/MitMaro/git-interactive-rebase-tool/) to help with this workflow.  It's an interactive editor for the `git rebase -i` todo-list.

bgilbert:
This will need to be rebased onto `main` to pick up #442.

config/common/errors.go
Adam0Brien: for the error name we could do : "ErrInvalidPartitionConfig" or is that too general?

config/fcos/v1_5_exp/validate_test.go
Adam0Brien: im getting this output when i run the test, its odd through becasue the other tests dont fail when an error message is in the err: field.

``` --- FAIL: TestValidateConfig (0.00s)
    --- FAIL: TestValidateConfig/validate_2 (0.00s)
        /home/adamobrien/coreos/coreos-butane/butane/config/fcos/v1_5_exp/validate_test.go:329: 
            	Error Trace:	/home/adamobrien/coreos/coreos-butane/butane/config/fcos/v1_5_exp/validate_test.go:329
            	Error:      	Not equal: 
            	            	expected: report.Report{Entries:[]report.Entry{report.Entry{Kind:0, Message:"unexpected partition number. Partition number should only be zero when using root disk or wipe_tables is true. Specify root disk as /dev/disk/by-id/coreos-boot-disk ", Context:path.ContextPath{Path:[]interface {}(nil), Tag:"yaml"}, Marker:tree.Marker{StartP:(*tree.Pos)(nil), EndP:(*tree.Pos)(nil)}}}}
            	            	actual  : report.Report{Entries:[]report.Entry{report.Entry{Kind:1, Message:"unexpected partition number. Partition number should only be zero when using root disk or wipe_tables is true. Specify root disk as /dev/disk/by-id/coreos-boot-disk ", Context:path.ContextPath{Path:[]interface {}{0}, Tag:"partition_number"}, Marker:tree.Marker{StartP:(*tree.Pos)(nil), EndP:(*tree.Pos)(nil)}}}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -3,7 +3,9 @@
            	            	   (report.Entry) {
            	            	-   Kind: (report.Kind) 0,
            	            	+   Kind: (report.Kind) 1,
            	            	    Message: (string) (len=165) "unexpected partition number. Partition number should only be zero when using root disk or wipe_tables is true. Specify root disk as /dev/disk/by-id/coreos-boot-disk ",
            	            	    Context: (path.ContextPath) {
            	            	-    Path: ([]interface {}) <nil>,
            	            	-    Tag: (string) (len=4) "yaml"
            	            	+    Path: ([]interface {}) (len=1) {
            	            	+     (int) 0
            	            	+    },
            	            	+    Tag: (string) (len=16) "partition_number"
            	            	    },
            	Test:       	TestValidateConfig/validate_2
            	Messages:   	invalid report
FAIL
FAIL	github.com/coreos/butane/config/fcos/v1_5_exp	0.006s
FAIL
 ```

**Comment by Adam0Brien** [comment]
Thanks for the help @prestist !!


> This will need to be rebased onto `main` to pick up #442.

that's no problem at all ill try and get that last test working!

File: config/fcos/v1_5_exp/validate_test.go [hunk 14]
```diff
@@ -317,15 +317,15 @@ func TestValidateConfig(t *testing.T) {
...
 
 	for i, test := range tests {
 		t.Run(fmt.Sprintf("validate %d", i), func(t *testing.T) {
 			actual := test.config.Validate()
 			expected := report.Report{}
-			expected.AddOnError(test.errPath, test.err)
+			expected.AddOnWarn(test.errPath, test.err)
 			assert.Equal(t, expected, actual, "invalid report")
 		})
 	}
```
Adam0Brien: this seems to fix it, makes sense as the 0th partition is the one that causes the fail.

config/fcos/v1_5_exp/validate_test.go
prestist: I saw that it looked like you already fixed the test. But as you now know the Context.Path you gave expected 'yaml' but the code returned 'partition_number'.

config/fcos/v1_5_exp/validate_test.go
prestist: Good job finding it!

prestist:
@Adam0Brien good job cleaning up the commits, I think we could maybe compress them into one commit in this case. Good job on the wording as well, def a big improvement!

docs/release-notes.md
```diff
@@ -45,6 +45,7 @@ nav_order: 9
 - Drop `LuksOption` and `RaidOption` types _(Go API for fcos 1.5.0+,
   flatcar 1.1.0+, openshift 4.14.0-experimental)_
 - Require Go 1.18+
+- test: Add a test for fcos/v1_6_exp to detect bad partition table configurations when wipe_tables is false and the partition table is not on the root disk.
```
prestist, line 48: Firstly; great attempt! 
So with our release notes (similarly to our commit messages) there are a few things to consider. 

You want to be concise as this is going to be a bulleted list at release time. (i.e https://github.com/coreos/butane/releases)

You also want to ask what part of the added work impacts the end user, if any (non developer)?
In this work example, the end user would be impacted more by the warning as they could see it. As the tests would only directly effect our developers working on the project. 

So we might want to try another swing at this with the context that this is a broadcast the the end user, and they would be reading this for what functionality was gained or lost since the last release.

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,19 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, disk := range c.Storage.Disks {
+		if disk.Device == "/dev/disk/by-id/coreos-boot-disk" {
+			continue
+		}
+		if disk.WipeTable != nil && !*disk.WipeTable {
```
prestist, line 60: We have a package called "util" which does nil check for ya, it should look like =>
`util.IsFalse(disk.WipeTable)`

config/common/errors.go
```diff
@@ -54,7 +54,8 @@ var (
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
 	// partition
-	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrInvalidPartitionConfig = errors.New("unexpected partition number. Partition number should only be zero when using root disk or wipe_tables is true. Specify root disk as /dev/disk/by-id/coreos-boot-disk ")
```
prestist, line 57: Sorry not trying to bike-shed, wdyt of this ? => 
"unexpected partition number; zero is for root disk or when wipe_tables is set to true. To declare a root disk specify the config.disk.device as /dev/disk/by-id/coreos-boot-disk")"

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,129 @@ func TestValidateGrubUser(t *testing.T) {
...
+
+func TestValidateConfig(t *testing.T) {
+	tests := []struct {
+		config  Config
+		err     error
+		errPath path.ContextPath
+	}{
+		// valid config
+		{
+			config: Config{
```
prestist, line 243: If you take a closer look to the test's above, it might make sense to follow a similar format. 

_Hint: look at how the elements in the array are initialized._

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,129 @@ func TestValidateGrubUser(t *testing.T) {
...
+										Number: 2,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			err:     nil,
+			errPath: path.New("yaml"),
```
prestist, line 266: If the `err` is nil we might want to consider dropping both `err` and `errPath` ?

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,129 @@ func TestValidateGrubUser(t *testing.T) {
...
+		},
+		// invalid config
```
prestist, line 322: Whats invalid about it ?

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,129 @@ func TestValidateGrubUser(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			err:     nil,
+			errPath: path.New("yaml"),
+		},
+		// specify root disk
```
prestist, line 295: maybe clarify this a bit more

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,19 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, disk := range c.Storage.Disks {
+		if disk.Device == "/dev/disk/by-id/coreos-boot-disk" {
```
prestist, line 57: Hmm, maybe we could join the two if blocks? 
If the device is not root, and wipe table is set to false, lets validate.

Adam0Brien: no problem! , my thinking was you could check if its the root disk and if it is then skip the rest of the test.

config/common/errors.go
Adam0Brien: sounds good ill add it now!

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: didn't know you could do that, have it added now though!

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: the config below had wipe_tables set to false and the 0th partition number set to 0. so i just commented it as an example of an invalid config

prestist:
Just noticed, when addressing the MR comments make sure to do the workflow described https://github.com/coreos/butane/pull/436#issuecomment-1478492293

As you want to keep the commit structure you had before the MR revisions.

config/fcos/v1_6_exp/validate.go
```diff
@@ -54,13 +54,12 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
...
 func (c Config) Validate() (r report.Report) {
 	for _, disk := range c.Storage.Disks {
-		if disk.Device == "/dev/disk/by-id/coreos-boot-disk" {
-			continue
-		}
-		if disk.WipeTable != nil && !*disk.WipeTable {
-			for p, partition := range disk.Partitions {
-				if partition.Number == 0 {
-					r.AddOnWarn(path.New("partition_number", p), common.ErrInvalidPartitionConfig)
+		if disk.Device != "/dev/disk/by-id/coreos-boot-disk" {
```
prestist, line 57: Hmm I was thinking more along the lines of 
```go
rootDevice := "/dev/disk/by-id/coreos-boot-disk"
if disk.Device != rootDevice && util.IsFalse(disk.WipeTable) {
//do work
}
```

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,13 +234,13 @@ func TestValidateGrubUser(t *testing.T) {
...
-		config  Config
-		err     error
+		in      Config
+		out     error
 		errPath path.ContextPath
 	}{
 		// valid config
 		{
-			config: Config{
+			in: Config{
```
prestist, line 243: Good observation! 'in' and 'out' make way more sense!

One last thing if you look at the _TestReportCorrleation_ you can see in the array it did not actually declare the individual fields it was defining.

You don't need to change over to this, Just wanted to make sure that you knew that you can do this.

prestist (commented):
Def need to move this commit into the prior one as a fixup.

config/fcos/v1_6_exp/validate.go
Adam0Brien: I see what you meant now my bad! ill add that in my next push.

docs/release-notes.md
Adam0Brien: pushed a rewording of it lmk what you think!

docs/release-notes.md
```diff
@@ -44,9 +44,9 @@ nav_order: 9
 - Drop `extensions` section _(fcos 1.5.0+, openshift 4.13.0+)_
 - Drop `LuksOption` and `RaidOption` types _(Go API for fcos 1.5.0+,
   flatcar 1.1.0+, openshift 4.14.0-experimental)_
-- Require Go 1.18+
-- test: Add a test for fcos/v1_6_exp to detect bad partition table configurations when wipe_tables is false and the partition table is not on the root disk.
-
+- Require Go 1.18+ 
+- Added warning for bad partition table configurations when a non-root partition tables wipe_tables is false and the partition number is zero in fcos/v1_6_exp.
```
prestist, line 48: Hmm, I would prefer to keep this at one line if we can, lets try and make it concise.

Additionally lets also look at the tense of the wording, the release notes is present tense. 

And when referencing configs you can do so like _(fcos 1.6.0-exp)_

prestist:
@Adam0Brien when you can, please squash your commits into one.

docs/release-notes.md
```diff
@@ -44,8 +44,9 @@ nav_order: 9
 - Drop `extensions` section _(fcos 1.5.0+, openshift 4.13.0+)_
 - Drop `LuksOption` and `RaidOption` types _(Go API for fcos 1.5.0+,
   flatcar 1.1.0+, openshift 4.14.0-experimental)_
-- Require Go 1.18+
-
+- Require Go 1.18+ 
+- Add warning for bad non-root partition tables when 'wipe_tables' is false and partition number is zero in _(fcos 1.6.0-exp)_
```
prestist, line 48: hmm, how about 

"Warn if specifying a partition number of zero on a non root device with `wipe_tables` set to false \_(fcos 1.6.0-exp)\_"

prestist: Actually, I think I changed my mind.
WDYT
"Warn on invalid partition configuration \_(fcos 1.6.0-exp)\_"

Adam0Brien: "Warn on invalid partition configuration _(fcos 1.6.0-exp)_" sounds good :+1:

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,120 @@ func TestValidateGrubUser(t *testing.T) {
...
+										Number: 0,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{ // This config specifies the root disk as /dev/disk/by-id/coreos-boot-disk (Valid Config)
```
prestist, line 290: small nit: all the other comments seem to be lower case.

prestist (commented):
Otherwise LGTM, 

Poke @bgilbert

config/fcos/v1_6_exp/validate_test.go
prestist: And lets not put it after the '{' , instead put it before the '{' on the line above.

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,17 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
...
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, disk := range c.Storage.Disks {
+		rootDevice := "/dev/disk/by-id/coreos-boot-disk"
+		if disk.Device != rootDevice && util.IsFalse(disk.WipeTable) {
+			for p, partition := range disk.Partitions {
+				if partition.Number == 0 {
+					r.AddOnWarn(path.New("partition_number", p), common.ErrInvalidPartitionConfig)
```
bgilbert, line 61: This isn't a valid context path.  It turns out that tests don't check for this at all, so I've added that in https://github.com/coreos/butane/pull/458.  Once that PR lands, please rebase this PR on top and add the `VerifyReport` call in the test function.

config/common/errors.go
```diff
@@ -54,7 +54,8 @@ var (
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
 	// partition
-	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrInvalidPartitionConfig = errors.New("unexpected partition number; zero is for root disk or when wipe_tables is set to true. To declare a root disk specify the config.disk.device as /dev/disk/by-id/coreos-boot-disk")
```
bgilbert, line 57: The first sentence seems to imply that, on the root disk or with `wipe_tables` true, the partition number _should_ be zero.  That's not correct.  Users can set the partition number to zero if they want, but not on non-root disks with `wipe_tables` false.  Could you make another try at this, focusing on specifically why the configuration isn't valid?

Also, instead of `config.disk.device`, `device` is probably clear enough, since the error will be attached to the adjacent partition number field.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,123 @@ func TestValidateGrubUser(t *testing.T) {
...
+										Number: 2,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		// invalid config, wipetables is false and the partition number is 0
```
bgilbert, line 316: Maybe something like:

```suggestion
		// invalid config (wipe_table is false with a partition numbered 0)
```

bgilbert (commented):
Some things to fix, but the general structure looks good!

bgilbert:
Oh, and please update the PR title/commit title/commit message to describe specifically what the change is.

config/common/errors.go
Adam0Brien: have it changed to "invalid partition configuration; Partition number should not be zero, except the root disk or when wipe_table is set to true. To declare a root disk specify the device as /dev/disk/by-id/coreos-boot-disk"

wdyt?

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,123 @@ func TestValidateGrubUser(t *testing.T) {
...
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device:    "/dev/disk/by-id/coreos-boot-disk",
+								WipeTable: util.BoolToPtr(false),
+								Partitions: []base.Partition{
+									{
+										Label:  util.StrToPtr("foo"),
+										Number: 1,
```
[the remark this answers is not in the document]
bgilbert: Yup!

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,17 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, disk := range c.Storage.Disks {
+		rootDevice := "/dev/disk/by-id/coreos-boot-disk"
```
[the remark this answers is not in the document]
Adam0Brien: sounds good, changes pushed!

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,95 @@ func TestValidateGrubUser(t *testing.T) {
...
+							{
+								Device:    rootDevice,
+								WipeTable: util.BoolToPtr(false),
+								Partitions: []base.Partition{
+									{
+										Label: util.StrToPtr("foo"),
+									},
+									{
+										Label:  util.StrToPtr("bar"),
+										Number: *util.IntToPtr(1),
```
bgilbert, line 282: `*util.IntToPtr(1)` is just `1`.

config/fcos/v1_6_exp/validate.go
```diff
@@ -22,6 +22,21 @@ import (
...
 )
 
+const rootDevice = "/dev/disk/by-id/coreos-boot-disk"
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for i, disk := range conf.Storage.Disks {
+		if disk.Device != rootDevice && (util.IsFalse(disk.WipeTable) || disk.WipeTable == nil) {
+			for p, partition := range disk.Partitions {
+				if partition.Number == 0 {
+					r.AddOnWarn(c.Append("storage", "disks", i, "partitions", p), common.ErrInvalidPartitionConfig)
```
bgilbert, line 32: Since we're warning specifically about the `number` field, let's add it to the path rather than pointing generically to the entire partition section.

config/common/errors.go
bgilbert: That fixes the "should be zero" problem, but still doesn't explain _why_ the configuration isn't okay.  Please take another look at #377 and try to summarize.  I can help edit the result.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,123 @@ func TestValidateGrubUser(t *testing.T) {
...
+	}{
+		// valid config
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device:    "/dev/vda",
+								WipeTable: util.BoolToPtr(false),
```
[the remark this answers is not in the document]
bgilbert: This is still pending.

bgilbert (commented):
The PR title and commit message still need details.

config/fcos/v1_6_exp/validate.go
```diff
@@ -22,6 +22,21 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
+const rootDevice = "/dev/disk/by-id/coreos-boot-disk"
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for i, disk := range conf.Storage.Disks {
+		if disk.Device != rootDevice && (util.IsFalse(disk.WipeTable) || disk.WipeTable == nil) {
```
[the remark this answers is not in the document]
Adam0Brien: didn't catch that thanks! :+1:

config/fcos/v1_6_exp/validate.go
Adam0Brien: Good point have that pushed now!

docs/release-notes.md
```diff
@@ -12,6 +12,8 @@ nav_order: 9
 ### Features
 
 - Add `-c`/`--check` option to check config without producing output
+- Improve partition validation logic for _(fcos 1.6.0-exp)_ to warn against invalid configurations.
+  Warns when wipe_table: false and any partitions have number: 0, except for root disk as "/dev/disk/by-id/coreos-boot-disk".
```
bgilbert, line 16: How about just:

```suggestion
- Warn if config attempts to reuse partition by label _(fcos 1.6.0-exp)_
```

config/common/errors.go
```diff
@@ -57,7 +57,8 @@ var (
 	ErrTooFewMirrorDevices     = errors.New("mirroring requires at least two devices")
 
 	// partition
-	ErrWrongPartitionNumber = errors.New("incorrect partition number; a new partition will be created using reserved label")
+	ErrInvalidPartitionConfig = errors.New("invalid partition configuration; Ignition will create a new partition each time the system is reprovisioned, so partition numbers must not be zero, except for the root disk or when wipe_table is set to true. To declare a root disk, specify the device as /dev/disk/by-id/coreos-boot-disk")
```
bgilbert, line 60: How about this:

```suggestion
	ErrReuseByLabel = errors.New("partitions cannot be reused by label; number must be specified except on boot disk (/dev/disk/by-id/coreos-boot-disk) or when wipe_table is true")
```

bgilbert (commented):
Tests are failing.

bgilbert:
For the commit message, I wrote this text:

```
fcos/v1_6_exp: warn if config attempts to reuse partition by label

Ignition matches existing partitions by number, not by label.  When
reusing an existing partition table, if the config specifies a label
and leaves the number at the default of 0, Ignition will try to add a
new partition rather than reusing any existing one.  That was probably
not the user's intention, so warn in this case.

However, the boot disk is a common exception, since it has a fresh
partition table on every run.  On FCOS we can detect the alias
/dev/disk/by-id/coreos-boot-disk and not warn, but that alias was added
recently and older configs won't use it.  On other variants we don't
have any way to detect the boot disk.  Therefore, don't implement the
warning in base and don't make it retroactive to existing stable specs;
only add it to the FCOS experimental spec.
```

Suggestions/edits welcome.

config/fcos/v1_6_exp/validate.go
```diff
@@ -22,6 +22,23 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
+const rootDevice = "/dev/disk/by-id/coreos-boot-disk"
+
+// Validate() operates on the top-level Config rather than Disk because
+// it needs to check all partitions in all disks, not just one disk
```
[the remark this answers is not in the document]
Adam0Brien: What do you think of 
"// Validate() operates on the top-level Config because the Disk struct is defined in base.Config"?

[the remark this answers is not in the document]
bgilbert: That's a good start but I'd probably be a bit more explicit.  Maybe e.g. "We can't define a Validate function directly on Disk because that's defined in base, so we use a Validate function on the top-level Config instead."

config/fcos/v1_6_exp/validate_test.go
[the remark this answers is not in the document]
bgilbert: It looks as though you removed an entire test instead of fixing `WipeTable`?

docs/release-notes.md
```diff
@@ -15,6 +15,7 @@ key](https://getfedora.org/security/).
 ### Features
 
 - Add `-c`/`--check` option to check config without producing output
+- Warn if config attempts to reuse partition by label _(fcos 1.6.0-exp)_
```
prestist, line 18: "Warn when a non root disk attempts to reuse partition by label _\_(fcos 1.6.0-exp)\__" WDYT?

Adam0Brien: Yeah i like that, it it too specific if we said "Unless wipe_table is true warn when a non root disk attempts to reuse partition by label _(fcos 1.6.0-exp)_"

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +236,110 @@ func TestValidateGrubUser(t *testing.T) {
...
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device:    rootDevice,
+								WipeTable: util.BoolToPtr(false),
+								Partitions: []base.Partition{
+									{
+										Label:  util.StrToPtr("bar"),
+										Number: 1,
```
bgilbert, line 278: We're not actually testing that `rootDevice` is special-cased, since the partition has `Number: 1`.  Let's drop this line.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +236,110 @@ func TestValidateGrubUser(t *testing.T) {
...
+		},
+		// invalid config (wipe_table is nil)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device:    "/dev/vda",
+								WipeTable: nil,
```
bgilbert, line 295: Explicit nil is redundant; drop.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +236,110 @@ func TestValidateGrubUser(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			out:     common.ErrReuseByLabel,
+			errPath: path.New("yaml", "storage", "disks", 0, "partitions", 0, "number"),
+		},
+		// invalid config (wipe_table is false with a partition numbered 0)
```
bgilbert, line 309: Did you mean to have an explicit `WipeTable: util.BoolToPtr(false)`?

bgilbert (commented):
Looks good overall!  A couple small fixes and then this should be ready.

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: I did, i thought that it was false by default but it makes more sense to keep it in thanks!

config/fcos/v1_6_exp/validate_test.go
bgilbert: To be clear, I was asking you to drop the line, not the entire test.

config/fcos/v1_6_exp/validate_test.go
bgilbert: Yeah, `bool` is false by default, but `*bool` is nil by default, and we want to check how we behave with `&false`.  Fix looks correct, thanks.

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: My bad will add it back now!

bgilbert:
LGTM.  Please squash your commits and I'll approve.

bgilbert (approved):

PR #439, opened 2023-03-13
docs/config-openshift-*: document location of SSH keys in RHEL 9

bgilbert:
OpenShift 4.13+ puts SSH keys in `.ssh/authorized_keys.d/ignition`.  Switch the 4.13 spec doc to that path.  For older spec docs we need to mention both paths, since Ignition will use the new path on 4.13+ even if an older Butane spec is used.


prestist (approved):
LGTM, thanks for updating the docs!

PR #440, opened 2023-03-14
docs/upgrading-openshift: document `password_hash` field in 4.13

bgilbert:


prestist (approved):
Oh;  
For future reference at what point in the process does this make more sense to be completed?
When the feature is added?
When stabilizing?

LGTM!

bgilbert:
It's currently a step in the spec stabilization checklist.  However, #289 tried adding commented-out entries for the upcoming stabilization, and that might be worth trying in the future, for the same reason that we've switched to maintaining release notes as we go.

PR #441, opened 2023-03-14
Fixes/cleanups for SSH key/systemd unit local file embedding

bgilbert:
Address various small items.  Follows up on #289.


base/v0_5_exp/translate_test.go
```diff
@@ -1935,257 +1935,158 @@ func TestTranslateSSHAuthorizedKey(t *testing.T) {
...
 			sshKeyDir,
 		},
 		{
-			"valid ssh_keys_local with multiple keys per file",
+			"valid local keys with multiple keys per file",
 			PasswdUser{SSHAuthorizedKeysLocal: []string{sshKeyMultipleKeysFileName}},
 			types.PasswdUser{SSHAuthorizedKeys: []types.SSHAuthorizedKey{
 				types.SSHAuthorizedKey(sshKey2),
 				types.SSHAuthorizedKey("#comment"),
-				types.SSHAuthorizedKey(""),
```
travier, line 1977: Aren't those empty lines here to make sure empty lines are correctly ignored?

travier:
Otherwise LGTM.

base/v0_5_exp/translate_test.go
travier: Maybe we should have a dedicated test for that case instead?

travier (approved):

prestist (approved):
Wow, nice fixup! 
LGTM

Issue #179, opened 2020-12-21
Add local file embedding for SSH keys, systemd units and dropins


travier:
With https://github.com/coreos/fcct/pull/99, we can use the following FCC to embed files from a directory in a config:
```
storage:
  files:
    contents:
      local: foo.conf
```

I would like to use a similar mechanism to embed:
  * SSH Keys
  * systemd units
  * systemd dropins

jlebon:
Not sure if this came up before, though some YAML parsers allow support for custom tags like `!include: myfile` which would solve this at a more generic level.

PR #442, opened 2023-03-15
fcos|r4e|flatcar spec stabilization

prestist:
https://github.com/coreos/butane/issues/434

PR #444, opened 2023-03-19
docs: add simple examples for SSH key & systemd file references (#179)

Okeanos:
Add minimal examples to the butane documentation explaining how to use file references instead of inlining SSH keys and systemd units introduced with #179 .

Noticed this was missing while working on [coreos-docs#511](https://github.com/coreos/fedora-coreos-docs/issues/511).

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -12,7 +12,7 @@ Here you can find a bunch of simple examples for using Butane, with some explana
 
 ## Users and groups
 
-This example modifies the existing `core` user and sets its ssh key.
+This example modifies the existing `core` user and sets its SSH key by providing it inline.
 
 <!-- butane-config -->
 ```yaml
@@ -25,7 +25,7 @@ passwd:
         - key1
 ```
 
-This example creates one user, `user1` and sets up one ssh public key for the user. The user is also given the home directory `/home/user1`, but it's not created, the user is added to the `wheel` and `plugdev` groups, and the user's shell is set to `/bin/bash`.
+This example creates one user, `user1` and sets up one SSH public key for the user. The user is also given the home directory `/home/user1`, but it's not created, the user is added to the `wheel` and `plugdev` groups, and the user's shell is set to `/bin/bash`.
 
 <!-- butane-config -->
 ```yaml
@@ -77,6 +77,24 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
 The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
 
 For more information, see the Fedora CoreOS documentation on [Authentication][fcos-auth-docs].
+
+### SSH keys from local files
+
+As shown in the previous examples you can inline multiple SSH public keys per user directly in the Butane config. Additionally, you can embed keys from local files at transpile time. This example creates a `core` user and configures its SSH keys from local files. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
+
+<!-- butane-config -->
+```yaml
+variant: fcos
+version: 1.5.0
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys_local:
+        - id_rsa.pub
+        - id_ed25519.pub
+```
+
+Different combinations for providing SSH keys are possible. You can provide inline ones together with file references and the files may also contain multiple keys (one per line). However, keep in mind that overall the keys must be unique. Check the [configuration specification][spec] for details.
 
 ## Storage and files
 
@@ -352,7 +370,7 @@ systemd:
             ExecStart=-/usr/sbin/agetty --autologin core --noclear %I $TERM
 ```
 
-This example creates a new systemd unit called hello.service, enables it so it will run on boot, and defines the contents to simply echo `"Hello, World!"`.
+This example creates a new systemd unit called `hello.service`, enables it so it will run on boot, and defines the contents to simply echo `"Hello, World!"`.
 
 <!-- butane-config -->
 ```yaml
@@ -372,6 +390,23 @@ systemd:
         [Install]
         WantedBy=multi-user.target
 ```
+
+This example specifies a systemd unit (`example.service`) and a dropin (`proxy.conf`) to be read from local files at transpile time. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
+
+<!-- butane-config -->
+```yaml
+variant: fcos
+version: 1.5.0
+systemd:
+  units:
+    - name: example.service
+      contents_local: example.service
+    - name: rpm-ostreed.service
+      dropins:
+          - name: proxy.conf
+            contents_local: example.conf
+```
+
 ## GRUB password
 
 This example adds a superuser to GRUB and sets a password. Users without the given username and password will not be able to access GRUB command line, modify kernel command-line arguments, or boot non-default OSTree deployments. Password hashes can be generated with `grub2-mkpasswd-pbkdf2`.
--- a/test
+++ b/test
@@ -52,6 +52,7 @@ if [ -n "${csplit}" ] && [ -n "${head}" ]; then
     mkdir -p tmpdocs/files-dir/tree
     touch tmpdocs/files-dir/{config.ign,ca.pem,example.conf,example.service,file,file-epilogue,local-file3}
     echo "ssh-rsa AAAA" > tmpdocs/files-dir/id_rsa.pub
+    echo "ssh-ed25519 AAAA" > tmpdocs/files-dir/id_ed25519.pub
 
     for doc in docs/*md
     do


bgilbert:
Requires #442.

bgilbert:
#442 merged, so this can be rebased to fix CI.

Okeanos:
Rebased and fixed the `./test` execution with an additional test file declaration. If you want me to change that to use `example.conf` instead for example instead of introducing an additional file let me know.

The release notes obviously fail right now (because I see the changes as part of the #179 issue) but I can touch the file if you want me to.

docs/examples.md
```diff
@@ -76,6 +76,26 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
 
 The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
 
+In case inline definitions of SSH keys are undesirable you can use file references instead. This example creates three users (`core`, `jlebon`, `miabbott`) and sets up their respective SSH keys via file references instead of inlining them. This requires using `butane` with the `--files-dir` parameter to allow loading files from disk.
```
bgilbert, line 79: This is being put in the password authentication subsection where it doesn't really belong.  Maybe let's add a new subsection?

For content, maybe something like:

```suggestion
In addition to defining SSH keys directly in the config, you can embed keys from local files at transpile time. This example creates a `core` user and configures its SSH keys from local files. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
```

docs/examples.md
```diff
@@ -76,6 +76,26 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
...
+    - name: core
+      ssh_authorized_keys_local:
+        - users/core/id_rsa.pub
+    - name: jlebon
+      ssh_authorized_keys_local:
+        - users/jlebon/id_rsa.pub
+        - users/jlebon/id_ed25519.pub
+    - name: miabbott
+      ssh_authorized_keys_local:
+        - users/miabbott/id_rsa.pub
```
bgilbert, line 96: The extra users complicate the example and don't really add anything.  Maybe one user with two keyfiles?

docs/examples.md
```diff
@@ -372,6 +392,29 @@ systemd:
         [Install]
         WantedBy=multi-user.target
 ```
+
+This example is uses a file reference to embed a locally defined file (`proxy.conf`) multiple times into the resulting Ignition file during transpile time. This requires using `butane` with the `--files-dir` parameter to allow loading files from disk.
```
bgilbert, line 396: Maybe something like:

```suggestion
This example specifies a systemd unit (`example.service`) and a dropin (`proxy.conf`) to be read from local files at transpile time. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
```

docs/examples.md
```diff
@@ -372,6 +392,29 @@ systemd:
...
+        - name: 99-proxy.conf
+          contents_local: proxy.conf
+    - name: zincati.service
+      dropins:
+        - name: 99-proxy.conf
+          contents_local: proxy.conf
+    - name: rpm-ostree-countme.service
+      dropins:
+        - name: 99-proxy.conf
+          contents_local: proxy.conf
```
bgilbert, line 415: Again, this is fairly redundant.  Let's embed one service and one dropin.

bgilbert (commented):
Thanks!

docs/examples.md
Okeanos: You're absolutely right. It wasn't actually my intention to put it into the password section.

docs/examples.md
Okeanos: Yeah, I get what you mean. My intention here was to give an example that immediately shows one of the reasons why one would use file embeds instead of just inlining a systemd service/dropin.

Okeanos:
I hopefully addressed your feedback now. If there's anything else you'd like adapted (or I overlooked) please let me know and I'll fix it.

docs/examples.md
```diff
@@ -76,6 +76,24 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
 
 The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
 
+In case inline definitions of SSH keys are undesirable you can use file references instead. This example creates three users (`core`, `jlebon`, `miabbott`) and sets up their respective SSH keys via file references instead of inlining them. This requires using `butane` with the `--files-dir` parameter to allow loading files from disk.
```
bgilbert, line 79: I think this is left over from the previous draft?

bgilbert, line 79: Is this paragraph left over from the previous draft?

docs/examples.md
```diff
@@ -76,6 +76,24 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
...
+
+### Handling SSH public keys
```
bgilbert, line 81: "Handling SSH public keys" is a bit generic, since we've already introduced SSH key configuration.  Maybe "SSH keys from local files"?

docs/examples.md
```diff
@@ -76,6 +76,24 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
...
+version: 1.5.0
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys_local:
+        - id_rsa.pub
+        - id_ed25519.pub
+```
+
 For more information, see the Fedora CoreOS documentation on [Authentication][fcos-auth-docs].
```
bgilbert, line 97: It's a bit ambiguous, but from the Git history it seems this was intended to be part of the password auth section.

test
```diff
@@ -50,8 +50,9 @@ if [ -n "${csplit}" ] && [ -n "${head}" ]; then
     trap 'rm -r tmpdocs' EXIT
     # Create files-dir contents expected by configs
     mkdir -p tmpdocs/files-dir/tree
-    touch tmpdocs/files-dir/{config.ign,ca.pem,example.conf,example.service,file,file-epilogue,local-file3}
+    touch tmpdocs/files-dir/{config.ign,ca.pem,example.conf,example.service,file,file-epilogue,local-file3,proxy.conf}
```
bgilbert, line 53: `proxy.conf` isn't needed here anymore, since we're reading from `example.conf`.

docs/examples.md
```diff
@@ -372,6 +390,23 @@ systemd:
...
+```yaml
+variant: fcos
+version: 1.5.0
+systemd:
+  units:
+    - name: example.service
+      contents_local: example.service
+    - name: rpm-ostreed.service
+      dropins:
+          - name: proxy.conf
```
bgilbert, line 406: Maybe it'd be clearer to use `example.conf` here (and in the description paragraph) rather than splitting the difference on naming?

docs/examples.md
```diff
@@ -76,6 +76,24 @@ $y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
 
 The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
 
+In case inline definitions of SSH keys are undesirable you can use file references instead. This example creates three users (`core`, `jlebon`, `miabbott`) and sets up their respective SSH keys via file references instead of inlining them. This requires using `butane` with the `--files-dir` parameter to allow loading files from disk.
+
+### Handling SSH public keys
+
+As shown in the previous examples you can inline multiple SSH public keys per user directly in the config file. Additionally, you can embed keys from local files at transpile time. This example creates a `core` user and configures its SSH keys from local files. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
```
bgilbert, line 83:
```suggestion
As shown in the previous examples you can inline multiple SSH public keys per user directly in the Butane config. Additionally, you can embed keys from local files at transpile time. This example creates a `core` user and configures its SSH keys from local files. The file paths are relative to the directory specified with the `--files-dir` command-line option, which must be provided.
```

bgilbert (commented):
Looks good!  Just a few nits.

docs/examples.md
Okeanos: Technically true, however, as pointed out in the [fcos-docs#511](https://github.com/coreos/fedora-coreos-docs/issues/511) issue and related PR (I should have another look at 🙈) the authentication page would be a good place for additional / more complex examples for SSH public key handling as well.

I'll move it back to the password settings section anyway.

docs/examples.md
Okeanos: Personally, I like the approach of having a file reference differently named from the target name of the unit we modify because it shows multiple things at once:

- how to modify a unit via dropins
- how file referencing works to provide unit/config contents
- that the file reference itself (its path/name) has _nothing_ to do with the actual unit that will be put into the Ignition config, i.e. it's up to the user how to arrange and name their local files

I like slightly more verbose examples that do not easily allow one to jump to conclusions such as "oh they are called the same so … I guess that's what I have to do to make it work". If you think this verbose approach to be unwarranted (in this example) I'll remove it, no worries.

bgilbert: Yeah, that makes sense.  I was partially reacting to the inconsistency: the dropin file uses a different name but the service doesn't.  Was that an oversight, or are you specifically looking to show that the local filename may or may not match the target filename?

Okeanos: Effectively the latter I suppose, i.e. the local file name may or may not match, however, with a single example this is difficult to show (as our discussion shows) because of perceived inconsistencies in the example itself. I have no really strong feelings one way or another and I think that users will figure it out correctly anyway. It's not rocket science after all :)

bgilbert (approved):
Looks great, thanks for handling this!

Issue #240, opened 2021-05-16
Fail if mountpoint path is forbidden by FCOS


rugk:
# Bug #

## Operating System Version ##

`fedora-coreos-34.20210503.1.1-live.x86_64` - current next stream

## Ignition Version ##

2.9.0

## Environment ##

bare-metal 

## Expected Behavior ##

Mounting succeeds.

## Actual Behavior ##

The last line in the log of `journalctl -t ignition`, even after printing the whole Ignition config is:
> }CRITICAL : Ignition failed: Mount path "/sysroot//mnt/external" contains non-directory component "/sysroot/mnt"

BTW yes the `}` at the start is really there, as it is the end of the Ignition config output it does there and in this case it seems to merge with that message…

This results in:
> ignition-complete.target: Job ignition-complete.target/start failed with result 'dependency'.

Doing an ls on `/sysroot/mnt`, I can see this is somewhat correct as it's a symbolic link:
```
$ ls -la /sysroot/mnt
[…] /sysroot/mnt -> var/mnt
```

As you can see, it links to `var/mnt` (_without_ a leading slash! – ~~also wonder whether that's intended~~).

~~In any case, in my emergency dracut shell I at least also cannot see any `/var/mnt` directory…~~

**Edit:** Ah of course it's intended by rpm-ostree's file layout. However, in the link's target, i.e. `/sysroot/var` there is also only a `lib` folder, no other folder, i.e. no `mnt` it actually expects. 

## Reproduction Steps ##

```yml
variant: fcos
version: 1.3.0
storage:
  luks:
    # external disk (already exists!)
    - name: external
      label: luks-external
      device: /dev/disk/by-id/[…]:0-part1
      key_file:
        inline: |-
          ********
      wipe_volume: false
  filesystems:
    - path: /mnt/external
      device: /dev/mapper/external
      format: btrfs
      wipe_filesystem: false
      with_mount_unit: true
 ```

BTW the `filesystems` -> systemd service for mounting the file is filesystem is correctly generated. 

## Other Information ##

I continue my challenge to mount an existing LUKS drive from https://github.com/coreos/ignition/issues/1210 ([originally here](https://discussion.fedoraproject.org/t/how-to-mount-re-use-an-existing-luks-volume/29767?u=rugk))

[...]

PR #449, opened 2023-03-30
fcos/v1_6_exp: Fail if mountpoint path is forbidden by FCOS

Adam0Brien:
Fix : #240

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -49,8 +49,9 @@ var (
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
 	// mount units
-	ErrMountUnitNoPath   = errors.New("path is required if with_mount_unit is true and format is not swap")
-	ErrMountUnitNoFormat = errors.New("format is required if with_mount_unit is true")
+	ErrMountUnitNoPath     = errors.New("path is required if with_mount_unit is true and format is not swap")
+	ErrMountUnitNoFormat   = errors.New("format is required if with_mount_unit is true")
+	ErrMountPointForbidden = errors.New("path must be under /etc or /var if with_mount_unit is true")
 
 	// boot device
 	ErrUnknownBootDeviceLayout = errors.New("layout must be one of: aarch64, ppc64le, x86_64")
--- a/config/fcos/v1_6_exp/validate.go
+++ b/config/fcos/v1_6_exp/validate.go
@@ -15,6 +15,8 @@
 package v1_6_exp
 
 import (
+	"regexp"
+
 	"github.com/coreos/butane/config/common"
 	"github.com/coreos/ignition/v2/config/util"
 
@@ -23,6 +25,8 @@ import (
 )
 
 const rootDevice = "/dev/disk/by-id/coreos-boot-disk"
+
+var allowedMountpoints = regexp.MustCompile(`^/(etc|var)(/|$)`)
 
 // We can't define a Validate function directly on Disk because that's defined in base,
 // so we use a Validate function on the top-level Config instead.
@@ -34,6 +38,11 @@ func (conf Config) Validate(c path.ContextPath) (r report.Report) {
 					r.AddOnWarn(c.Append("storage", "disks", i, "partitions", p, "number"), common.ErrReuseByLabel)
 				}
 			}
+		}
+	}
+	for i, fs := range conf.Storage.Filesystems {
+		if fs.Path != nil && !allowedMountpoints.MatchString(*fs.Path) && util.IsTrue(fs.WithMountUnit) {
+			r.AddOnError(c.Append("storage", "filesystems", i, "path"), common.ErrMountPointForbidden)
 		}
 	}
 	return
--- a/config/fcos/v1_6_exp/validate_test.go
+++ b/config/fcos/v1_6_exp/validate_test.go
@@ -237,6 +237,143 @@ func TestValidateGrubUser(t *testing.T) {
 	}
 }
 
+func TestValidateMountPoints(t *testing.T) {
+	tests := []struct {
+		in      Config
+		out     error
+		errPath path.ContextPath
+	}{
+		// valid config (has prefix "/etc" or "/var")
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/etc/foo"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+							{
+								Path:          util.StrToPtr("/var"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+							{
+								Path:          util.StrToPtr("/invalid/path"),
+								WithMountUnit: util.BoolToPtr(false),
+							},
+							{
+								WithMountUnit: util.BoolToPtr(true),
+							},
+							{
+								Path:          nil,
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+		},
+		// invalid config (path name is '/')
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+		// invalid config (path is /boot)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/boot"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+		// invalid config (path is invalid, does not contain /etc or /var)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/thisIsABugTest"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+		// invalid config (path is /varnish)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/varnish"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+		// invalid config (path is /foo/var)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:          util.StrToPtr("/foo/var"),
+								WithMountUnit: util.BoolToPtr(true),
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("validate %d", i), func(t *testing.T) {
+			actual := test.in.Validate(path.New("yaml"))
+			baseutil.VerifyReport(t, test.in, actual)
+			expected := report.Report{}
+			expected.AddOnError(test.errPath, test.out)
+			assert.Equal(t, expected, actual, "invalid report")
+		})
+	}
+}
+
 func TestValidateConfig(t *testing.T) {
 	tests := []struct {
 		in      Config
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -19,6 +19,8 @@ key](https://getfedora.org/security/).
 - Add `-c`/`--check` option to check config without producing output
 - Warn if config attempts to reuse partition by label _(fcos 1.6.0-exp,
   openshift 4.14.0-exp)_
+- Require `storage.filesystems.path` to start with `/etc` or `/var` if
+  `with_mount_unit` is true _(fcos 1.6.0-exp, openshift 4.14.0-exp)_
 
 ### Bug fixes
 


config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,23 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func isMountpointForbidden(path string) bool {
+	// Check if the path is a restricted directory
+	restrictedDirs := []string{"/boot", "/sys", "/usr", "/var", "/", "/sysroot/mnt"}
```
jobcespedes, line 57: Hi,

`/opt` could be added too

Adam0Brien: Hi thanks for the feedback!
Why should /opt be added?

Adam0Brien: makes sense thanks for letting me know!

config/fcos/v1_6_exp/validate.go
```diff
@@ -54,7 +54,7 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 
 func isMountpointForbidden(path string) bool {
 	// Check if the path is a restricted directory
-	restrictedDirs := []string{"/boot", "/sys", "/usr", "/var", "/", "/sysroot/mnt"}
+	restrictedDirs := []string{"/boot", "/sys", "/usr", "/var", "/", "/sysroot/mnt", "/opt"}
```
prestist, line 57: This commit looks like it should be apart of the original.

prestist: And wdyt about making these alphabetical?

config/common/errors.go
```diff
@@ -46,8 +46,9 @@ var (
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
 	// mount units
-	ErrMountUnitNoPath   = errors.New("path is required if with_mount_unit is true and format is not swap")
-	ErrMountUnitNoFormat = errors.New("format is required if with_mount_unit is true")
+	ErrMountUnitNoPath     = errors.New("path is required if with_mount_unit is true and format is not swap")
+	ErrMountUnitNoFormat   = errors.New("format is required if with_mount_unit is true")
+	ErrMountPointForbidden = errors.New("mount point is forbidden by fcos, use alternative mount point")
```
prestist, line 51: Hmm, it might make sense to put the mount points that are forbidden?

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,23 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
...
+	restrictedDirs := []string{"/boot", "/sys", "/usr", "/var", "/", "/sysroot/mnt"}
+	for _, dir := range restrictedDirs {
+		if path == dir {
+			return true
+		}
+	}
+	return false
+}
+
+func (c Config) ValidateMountPoints() (r report.Report) {
```
prestist, line 66: WDYT about calling this `validate()`?

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,23 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func isMountpointForbidden(path string) bool {
```
prestist, line 55: WDYT about moving this logic into `ValidateMountPoints()` function?

Since we are not currently calling this function but one time (for `ValidateMountPoints`) I wonder if its necessary to separate the logic?

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,23 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
...
+			return true
+		}
+	}
+	return false
+}
+
+func (c Config) ValidateMountPoints() (r report.Report) {
+	for m, mountPoint := range c.Storage.Filesystems {
+		if isMountpointForbidden(*mountPoint.Path) {
+			r.AddOnError(path.New("forbidden_mountpoint", m), common.ErrMountPointForbidden)
```
prestist, line 69: I think the correct path is more like `path.new("filesystems",i,"path")`

prestist: I like to think of the path as a map back to the problem area of the config for the end user to debug.

config/fcos/v1_6_exp/validate.go
prestist: Hmm, it might be less of a footprint to be less explicit here. What if we flip the logic on its head, and only allow the allowed paths and their subpaths?

Looking at the [docs](https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_configuration_in_etc_and_state_in_var) we only allow `/etc` and `/var`

docs/release-notes.md
```diff
@@ -45,6 +45,7 @@ nav_order: 9
 - Drop `LuksOption` and `RaidOption` types _(Go API for fcos 1.5.0+,
   flatcar 1.1.0+, openshift 4.14.0-experimental)_
 - Require Go 1.18+
+- Fail on invalid mointpoint configuration
```
prestist, line 48: Lol this scared me for a second, make sure to fetch upstream and update your branch so that you dont have to move around your release notes.

prestist: (thought I had not updated the release notes yet even though release had already happened)

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -231,3 +231,76 @@ func TestValidateGrubUser(t *testing.T) {
...
+	}{
+		// valid config
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path:           util.StrToPtr("/foo"),
+								WipeFilesystem: util.BoolToPtr(false),
```
prestist, line 249: Generally try and make your configs as lean as possible to be explicit about what you are testing. 
ie. `WipeFilesystem`

prestist (commented):
Thank you for working on this, its coming along good. Just a few comments.

config/fcos/v1_6_exp/validate.go
Adam0Brien: i was thinking about that but wouldn't that mess with [](https://github.com/coreos/butane/pull/436)?

Adam0Brien: because then there would be two validate methods that use config

config/fcos/v1_6_exp/validate.go
Adam0Brien: Good point, have that fixed now!

config/fcos/v1_6_exp/validate.go
Adam0Brien: Makes sense to do it that way, thanks!

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: not sure why i kept that in, have it fixed now!

config/fcos/v1_6_exp/validate.go
Adam0Brien: ill keep that in mind the next time i use this approach!

config/fcos/v1_6_exp/validate.go
prestist: That is a good question, you will have a conflict once one or the other merges, but that should not impact the naming, the conflict should be fairly easy to solve, and the validate() can include both sections you are validating

docs/release-notes.md
Adam0Brien: my bad should be fixed now!

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,13 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, fs := range c.Storage.Filesystems {
+		fsPath := *fs.Path
+		if len(fsPath) < 4 || (fsPath[:4] != "/etc" && fsPath[:4] != "/var") {
+			r.AddOnError(path.New("filesystems", 0), common.ErrMountPointForbidden)
```
prestist, line 59: hmm, we should report correct number here.

config/fcos/v1_6_exp/validate.go
```diff
@@ -51,3 +51,13 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (c Config) Validate() (r report.Report) {
+	for _, fs := range c.Storage.Filesystems {
+		fsPath := *fs.Path
+		if len(fsPath) < 4 || (fsPath[:4] != "/etc" && fsPath[:4] != "/var") {
```
prestist, line 58: Hmm, I think this works but what do you think about maybe using [regex](https://pkg.go.dev/regexp)?

config/common/errors.go
```diff
@@ -46,8 +46,9 @@ var (
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
 	// mount units
-	ErrMountUnitNoPath   = errors.New("path is required if with_mount_unit is true and format is not swap")
-	ErrMountUnitNoFormat = errors.New("format is required if with_mount_unit is true")
+	ErrMountUnitNoPath     = errors.New("path is required if with_mount_unit is true and format is not swap")
+	ErrMountUnitNoFormat   = errors.New("format is required if with_mount_unit is true")
+	ErrMountPointForbidden = errors.New("mount point is forbidden by fcos, use '/etc','/var' or their subdirectories ")
```
prestist, line 51: Hmm, what about "forbidden mount point. Use a mount point that starts with `/etc` or `/var`".

prestist (commented):
Some small finds.

config/fcos/v1_6_exp/validate.go
Adam0Brien: not sure why i didn't catch that my bad!

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
 
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
+func (c Config) Validate() (r report.Report) {
+	regex := regexp.MustCompile(`^/(etc|var)/.*`)
+	for _, fs := range c.Storage.Filesystems {
+		fsPath := *fs.Path
+		if len(fsPath) < 4 || !regex.MatchString(fsPath) {
```
prestist, line 31: Hmm, is the len(fsPath) still a necessary check? and I think it might be introducing a bug.

We might want some tests that cover the case of something that is not strictly '/' '/etc' '/var' '/boot' IE, what about a path like '/thisIsABugTest' ?

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,83 @@ func TestValidateGrubUser(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateMountPoints(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
Adam0Brien: didn't think about a null mount point, i've added a check and a test for that now!

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage"),
+		},
+		//invalid config (path is )
```
prestist, line 323: Nit: could we add a space '//' 'comment'

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage"),
+		},
+		//invalid config (path is /boot)
```
prestist, line 275: Nit: could we add a space '//' 'comment'

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage"),
+		},
+		//valid config (has prefix "/var")
```
prestist, line 292: Nit: could we add a space '//' 'comment'

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
...
+						Filesystems: []base.Filesystem{
+							{
+								Path: util.StrToPtr("/var/foo/bar"),
+							},
+						},
+					},
+				},
+			},
+		},
+		//invalid config (path is nil)
```
prestist, line 306: Nit: could we add a space '//' 'comment'

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
 
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
+var regex = regexp.MustCompile(`^/(etc|var)/.*`)
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for _, fs := range conf.Storage.Filesystems {
+		if (fs.Path == nil) || (!regex.MatchString(*fs.Path)) {
```
prestist, line 31: In the past how did butane handle a nil path?

prestist, line 31: Hmm, I am not a big fan of '!' logic, (total nit) but could we maybe try something like if it is good then we continue, otherwise error?

prestist (dismissed):
Small nits (WHOOOPS meant to comment not approve!)

config/fcos/v1_6_exp/validate.go
Adam0Brien: what do you think of this?
```
func (conf Config) Validate(c path.ContextPath) (r report.Report) {
	for _, fs := range conf.Storage.Filesystems {
		if util.NilOrEmpty(fs.Path) {
			r.AddOnError(c.Append("storage"), common.ErrMountPointForbidden)
		} else if (regex.MatchString(*fs.Path)){
			continue
		}else{
			r.AddOnError(c.Append("storage"), common.ErrMountPointForbidden)
		}
	}
	return
}
```

prestist: Eh nvm disregard.

config/common/errors.go
```diff
@@ -49,8 +49,9 @@ var (
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
 	// mount units
-	ErrMountUnitNoPath   = errors.New("path is required if with_mount_unit is true and format is not swap")
-	ErrMountUnitNoFormat = errors.New("format is required if with_mount_unit is true")
+	ErrMountUnitNoPath     = errors.New("path is required if with_mount_unit is true and format is not swap")
+	ErrMountUnitNoFormat   = errors.New("format is required if with_mount_unit is true")
+	ErrMountPointForbidden = errors.New("forbidden mount point. Use a mount point that starts with /etc or /var")
```
prestist, line 54: hmm if you look at all the other errors '.' tends to only be used as a decimal and not a period.

How about "forbidden mount point, use a mount point that starts with /etc or /var" ?

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,30 @@
...
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
+var regex = regexp.MustCompile(`^/(etc|var)/.*`)
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for _, fs := range conf.Storage.Filesystems {
+		if util.NilOrEmpty(fs.Path) {
+			r.AddOnError(c.Append("storage"), common.ErrMountPointForbidden)
```
prestist, line 32: I dont like having to have dupe code like these two "addOnErrror". I think we should step back to how it used to be. Sorry for the extra steps.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage"),
+		},
+		// invalid config (path is )
```
prestist, line 323: missing path

prestist (commented):
Small nits, otherwise lgtm

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: my bad ended up just calling it "path is too long" because of the previous code which checked if the length was over 4 in length, regex was a better way to go in the end!

config/fcos/v1_6_exp/validate.go
Adam0Brien: no problem have it changed back now!

config/common/errors.go
Adam0Brien: my bad didn't notice i used a '.' thanks!

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
+	"regexp"
+
 	"github.com/coreos/butane/config/common"
 	"github.com/coreos/ignition/v2/config/util"
 
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
+var regex = regexp.MustCompile(`^/(etc|var)/.*`)
```
bgilbert, line 27: Nit: the trailing `.*` is redundant.

Note https://github.com/coreos/butane/issues/240#issuecomment-1061873909.  @jlebon, are there other paths we should also support?  Note that we're currently failing, not warning, on objectionable paths.

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for _, fs := range conf.Storage.Filesystems {
+		if util.NilOrEmpty(fs.Path) || (!regex.MatchString(*fs.Path)) {
```
bgilbert, line 31: It's legal not to specify the `path`, if you want to format a filesystem but not mount it.

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
+			r.AddOnError(c.Append("storage"), common.ErrMountPointForbidden)
```
bgilbert, line 32: The path should point to the field we're complaining about, not generically to the top of the `storage` section.

docs/release-notes.md
```diff
@@ -63,6 +63,7 @@ key](https://getfedora.org/security/).
   3.5.0-experimental
 - Add RHEL for Edge spec 1.2.0-experimental, targeting Ignition spec
   3.5.0-experimental
+- Fail on invalid mointpoint configuration
```
bgilbert, line 66: Let's say specifically what we're forbidding, and which specs we're forbidding it in.

config/common/errors.go
```diff
@@ -49,8 +49,9 @@ var (
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
 	// mount units
-	ErrMountUnitNoPath   = errors.New("path is required if with_mount_unit is true and format is not swap")
-	ErrMountUnitNoFormat = errors.New("format is required if with_mount_unit is true")
+	ErrMountUnitNoPath     = errors.New("path is required if with_mount_unit is true and format is not swap")
+	ErrMountUnitNoFormat   = errors.New("format is required if with_mount_unit is true")
+	ErrMountPointForbidden = errors.New("forbidden mount point, Use a mount point that starts with /etc or /var")
```
bgilbert, line 54: This block is currently for mount unit generation (`with_mount_unit: true`), and none of the other blocks seem to fit.  Let's either start a new block or update the header comment for this one.

Maybe we could tighten up the error text e.g. like this?
```suggestion
	ErrMountPointForbidden = errors.New("mounts must be under /etc or /var")
```

docs/release-notes.md
bgilbert: Nope, it isn't.

config/fcos/v1_6_exp/validate.go
jlebon: Won't this regex also fail on just `/var`? We have a test for this [in FCOS](https://github.com/coreos/fedora-coreos-config/blob/52b1d1adfc7668bba1cc28bc14bd03a35de2ced4/tests/kola/var-mount/simple/config.bu#L17), but CoreOS CI doesn't run here.

Not sure on the use cases for mounting something under `/etc/` has, but agree it should work.

I think that should cover it otherwise! As long as we're applying this to just mount points (which it looks like that's the case), https://github.com/coreos/butane/issues/240#issuecomment-1061873909 is not a concern.

Adam0Brien: ill replace the regex with '/(etc|var)' that seems to work for both paths

config/fcos/v1_6_exp/validate.go
Adam0Brien: no problem its pointing at the path field now!

docs/release-notes.md
Adam0Brien: WDYT of "- Forbid relabeling mountpoints which already exist _(fcos 1.6-exp)_"?

config/fcos/v1_6_exp/validate.go
bgilbert: It also allows `/varnish`.  Also, since you've removed the anchor, it allows `/foo/var`.  How about `^/(etc|var)(/|$)` ?

config/fcos/v1_6_exp/validate.go
bgilbert: This still needs to be fixed.  The code currently fails if the path is unspecified.

docs/release-notes.md
bgilbert: Wrong PR.  :grin:  This one isn't related to relabeling.

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
 	"github.com/coreos/ignition/v2/config/util"
 
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
+var regex = regexp.MustCompile(`/(etc|var)`)
+
+func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	for fs, filesystem := range conf.Storage.Filesystems {
```
bgilbert: Nit: `i` and `fs` would be more idiomatic than `fs` and `filesystem`.

docs/release-notes.md
Adam0Brien: My bad i was looking at #436 beforehand :sweat_smile:

config/fcos/v1_6_exp/validate.go
```diff
@@ -15,13 +15,26 @@
...
+		if util.NilOrEmpty(filesystem.Path) || (!regex.MatchString(*filesystem.Path)) {
```
bgilbert: I just realized we need one more check here: we should only fail if `util.IsTrue(filesystem.WithMountUnit)`.  The semantics of `path` are:

> the mount-point of the filesystem while Ignition is running relative to where the root filesystem will be mounted. This is not necessarily the same as where it should be mounted in the real root, but it is encouraged to make it the same.

So, it's legal for the config to specify arbitrary `path`s for filesystems (which will only be used when Ignition is running) and then manually arrange for the filesystem to be mounted somewhere else.  However, most configs probably set `with_mount_unit: true` to have Butane arrange for the mounting of the filesystem, in which case the `path` does need to be a valid mountpoint.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -234,3 +234,117 @@ func TestValidateGrubUser(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateMountPoints(t *testing.T) {
+	tests := []struct {
+		in      Config
+		out     error
+		errPath path.ContextPath
+	}{
```
bgilbert, line 243: Per https://github.com/coreos/butane/pull/449#discussion_r1191432117, let's also check `/var`, `/varnish` and `/foo/var`.

config/common/errors.go
```diff
@@ -48,6 +48,9 @@ var (
 	// systemd
 	ErrTooManySystemdSources = errors.New("only one of the following can be set: contents, contents_local")
 
+	// mount points
+	ErrMountPointForbidden = errors.New("mounts must be under /etc or /var")
```
bgilbert, line 52: Due to https://github.com/coreos/butane/pull/449#discussion_r1191447738, this is actually a mount unit error after all, so let's move it back.  :flushed:

And we should probably clarify it a bit.  For consistency with the others, `path must be under /etc or /var if with_mount_unit is true`?

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: No problem, have them added now

config/common/errors.go
Adam0Brien: no problem, changes pushed!

docs/release-notes.md
Adam0Brien: Changed to "Forbid mountpoints that don't have the prefix /var or /etc if with_mount_unit is true _(fcos 1.6-exp)_"

bgilbert: How about this?

```
- Require `storage.filesystems.path` to start with `/etc` or `/var` if
  `with_mount_unit` is true _(fcos 1.6.0-exp, openshift 4.14.0-exp)_
```

Note hard line-wrap at 80 columns.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -237,6 +237,141 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+							{
+								Path: util.StrToPtr("/var"),
+							},
+						},
+					},
+				},
+			},
+		},
+		// invalid config (path name is '/')
```
bgilbert, line 263: We're not setting `out`, so this is actually testing that the config is valid.  Needs `WithMountUnit`.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -237,6 +237,141 @@ func TestValidateGrubUser(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+
+			out:     common.ErrMountPointForbidden,
+			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
+		},
+		// valid config (path is nil)
```
bgilbert, line 295: I think this can be dropped in favor of the expanded "valid" test case above.  As written, this one isn't especially useful, since it has *two* conditions that would cause us not to warn.

bgilbert (commented):
You'll need to rebase this to pick up the CI config changes in #471.

config/fcos/v1_6_exp/validate.go
```diff
@@ -24,6 +26,8 @@ import (
 
 const rootDevice = "/dev/disk/by-id/coreos-boot-disk"
 
+var regex = regexp.MustCompile(`^/(etc|var)(/|$)`)
```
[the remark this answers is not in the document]
Adam0Brien: how is `allowedMountpoints` or maybe `legalMountpoints`?

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: No probkem, made a separate commit with the dropped test

config/fcos/v1_6_exp/validate.go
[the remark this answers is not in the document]
bgilbert: `allowedMountpoints` works :+1:

config/fcos/v1_6_exp/validate_test.go
bgilbert: Not done.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -237,6 +237,134 @@ func TestValidateGrubUser(t *testing.T) {
...
+		},
+		// invalid config (path name is '/')
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path: util.StrToPtr("/"),
+							},
```
bgilbert, line 282: Missing `WithMountUnit: true`.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -237,6 +237,134 @@ func TestValidateGrubUser(t *testing.T) {
...
+		},
+		// valid config (path is nil)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path: nil,
+							},
```
bgilbert, line 296: Missing `WithMountUnit: true`.

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: My bad i've commited this will push when all changes are made:
```
// invalid config (path name is '/')
		{
			in: Config{
				Config: base.Config{
					Storage: base.Storage{
						Filesystems: []base.Filesystem{
							{
								Path:          util.StrToPtr("/"),
								WithMountUnit: util.BoolToPtr(true),
							},
						},
					},
				},
			},
			out:     common.ErrMountPointForbidden,
			errPath: path.New("yaml", "storage", "filesystems", 0, "path"),
		},
```
Ran tests and it passes also

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: Got confused about what part needed changing thanks for specifying 
Dropped //valid config (path is nil)

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -237,6 +237,141 @@ func TestValidateGrubUser(t *testing.T) {
...
+	}{
+		// valid config (has prefix "/etc" or "/var")
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Filesystems: []base.Filesystem{
+							{
+								Path: util.StrToPtr("/etc/foo"),
+							},
```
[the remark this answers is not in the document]
Adam0Brien: My bad switched WithMountUnit to true and removed the `out` and `errPath`

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -250,15 +250,26 @@ func TestValidateMountPoints(t *testing.T) {
...
+								WithMountUnit: util.BoolToPtr(false),
 							},
 							{
-								Path: util.StrToPtr("/var"),
+								Path:          util.StrToPtr("/var"),
+								WithMountUnit: util.BoolToPtr(false),
+							},
+							{
+								Path:          util.StrToPtr("invalid/path"),
+								WithMountUnit: util.BoolToPtr(true),
```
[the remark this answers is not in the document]
Adam0Brien: taken it out of this test case, will i make it its own separate test?

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: Added it to the (path name is '/') test

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: Have dropped the test

config/fcos/v1_6_exp/validate_test.go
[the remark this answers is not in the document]
Adam0Brien: Have swapped to false

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -250,15 +250,26 @@ func TestValidateMountPoints(t *testing.T) {
...
+							},
+							{
+								WithMountUnit: util.BoolToPtr(false),
```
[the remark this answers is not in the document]
Adam0Brien: Have swapped to True

config/fcos/v1_6_exp/validate_test.go
Adam0Brien: Have aded Withmountunit and moved to the first test 

		// valid config (has prefix "/etc" or "/var")

bgilbert (approved):
Looks good!

PR #452, opened 2023-04-05
internal: add `-c`/`--check` option to skip output

bgilbert:
Add option to validate config without producing Ignition or MachineConfig output.


bgilbert:
No, I was writing up #453 and wanted to be able to validate a config without redirecting to `/dev/null`.

PR #454, opened 2023-04-11
templates/stabilize-checklist: note spec stabilization in release notes

bgilbert:
Matches a similar checklist change in Ignition.


prestist (approved):
LGTM

PR #457, opened 2023-04-11
config: define error structs for errors that have context

bgilbert:
Butane has two sites that want to include additional information in an error string.  Rather than requiring callers to string-match to detect these, define structs with the error details.

This should not be a breaking change.


prestist (approved):
LGTM!

PR #458, opened 2023-04-12
tests: verify that report paths are valid

bgilbert:
A common mistake is to create `ContextPaths` that aren't valid references to an item in the config.  `DebugVerifyCoverage()` should catch such paths in a `TranslationSet`, but validation errors (during either the validation or translation passes) may add paths to a report without a corresponding `TranslationSet` entry.  Add helper functions to validate paths in a report, and use them in every test case.

In particular, call the helpers even in tests that expect empty reports, to reduce the risk that a test created in the future will omit the check.

Fix context paths for forbidden field errors in all versions of the `r4e` spec.

### DIFF
--- a/base/util/test.go
+++ b/base/util/test.go
@@ -15,9 +15,15 @@
 package util
 
 import (
+	"reflect"
+	"strings"
 	"testing"
 
+	confutil "github.com/coreos/butane/config/util"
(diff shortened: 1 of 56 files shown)


prestist (approved):
Freaking great! thank you for this automation, its going to keep paths healthy! 

LGTM

PR #460, opened 2023-04-20
Automatically generate spec docs

bgilbert:
Generate docs using Ignition's docs generator, and a combination of Ignition's config and our own overlay.  The `main.go` glue code is based on Ignition's, with some additional logic e.g. to convert everything to snake case.

Our `DocNode`s are merged like this: first, Butane's config is merged into Ignition's.  Any components that match a component in Ignition's config are merged directly with that component.  (This only applies to `resource` right now.)  Then, the root component is walked, and every component reference (`use:`) is merged in.  A consequence is that fields in Butane's `root` component that patch fields in Ignition's `resource` component need an `after` attribute, even though it looks spurious, because they're not merged with their parent field until the _second_ merge.

Merging two nodes is similar to Ignition config merging: a parent value is replaced with a child value if specified, and lists of values are joined (with the parent values coming first).  `required` and `required-if` override each other.  In the upstream/downstream merge, the Ignition components are the parents, and in the component/root merge, the component is the parent.


internal/doc/butane.yaml
```diff
@@ -0,0 +1,439 @@
...
+        - name: proxy
+          transforms:
+            # Snake-case cross-references between children
+            - regex: "`(https|no)Proxy`"
+              replacement: "`${1}_proxy`"
+              descendants: true
+    - name: storage
+      children:
+        - name: disks
+          unless:
```
bgilbert, line 193: There are 24 of these directives and they clutter things up.  I'd like to look at deriving this info automatically, but that'll require some refactoring and I'll save it for a followup.

prestist (approved):
LGTM, thank you for getting this in!

PR #463, opened 2023-05-01
getting-started: update link to fedora signing keys

### LINKED ISSUE #462: Wrong link to Fedora GPG Signing Key
The [getting started](https://github.com/coreos/butane/blob/main/docs/getting-started.md) docs say to download the Fedora signing key from `https://getfedora.org/static/fedora.gpg` to verify the releases – however, this url ends in a 404 right now.

According to the [Fedora CoreOS docs](https://docs.fedoraproject.org/en-US/fedora-coreos/producing-ign/#_standalone_binary) one should look it up on the [Security page](https://fedoraproject.org/security/) which links to `https://fedoraproject.org/fedora.gpg` instead.

prestist:
Update the docs link to latest location of fedora signing keys. fixes #462


bgilbert (approved):
Thanks.  Could you handle updating fedora-coreos-docs as well?

prestist:
added a issue for the change in fedora-coreos docs  https://github.com/coreos/fedora-coreos-docs/issues/547

PR #465, opened 2023-05-04
Really fix line/column reporting for `r4e` unsupported field errors

bgilbert:
We were expecting translation functions to return `yaml` paths but weren't ensuring it.  The `r4e` `checkForForbiddenFields()` functions weren't, causing their errors not to identify the correct line and column.

Since #458, `TranslateReportPaths()` automatically ignores paths that are already in `yaml` syntax, so we can just call it once from `config/util.Translate()` and stop requiring each individual translator to call it.

This is a breaking change for any external caller using the `*Unvalidated` functions, since those may now (legally) return a mix of `yaml` and `json` paths.  Probably those functions have no external callers.

While refactoring tests to support this, drop `VerifyTranslatedReport()`.  Tests now translate reports inline and use `VerifyReport()` directly.

### DIFF
--- a/base/util/test.go
+++ b/base/util/test.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 	"testing"
 
-	confutil "github.com/coreos/butane/config/util"
 	"github.com/coreos/butane/translate"
 
 	"github.com/coreos/vcontext/path"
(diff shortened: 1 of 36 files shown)


prestist (approved):
Good job, and nice quick iteration!

LGTM!

PR #466, opened 2023-05-06
base: properly snake-case `http_headers` translation

bgilbert:
Fix line/column reporting for `http_headers` errors.


prestist (approved):
LGTM

PR #473, opened 2023-06-07
internal/doc: document `/dev/disk/by-id/coreos-boot-disk` symlink

bgilbert:
It's available in current versions of FCOS and in OpenShift 4.11+.  Older FCOS releases don't support it, but Butane doesn't support _them_, so document it in every `fcos` spec.


prestist (approved):
Thank you for updating that;

LGTM

PR #476, opened 2023-07-03
base/v0_6_exp: Validate merged/replaced Ignition configs

### LINKED ISSUE #275: Validate merged/replaced Ignition configs if they're local/inline
When merging/replacing a `local` or `inline` config, we already have the config contents, so we should run it through Ignition config validation to catch problems early.  (For example, the user may have provided a Butane config by mistake.)

Adam0Brien:
Fix #275 

This PR introduces validation for local/inline fields when using the merge/replace feature in butane.

For this PR ignitions [Parse](https://github.com/coreos/ignition/blob/5a97eae4e831eb81f250def93ed64837aa45b50e/config/config.go#L26) function needs to be used so the vendor folder is has been updated to accommodate this

### DIFF
--- a/base/v0_6_exp/validate.go
+++ b/base/v0_6_exp/validate.go
@@ -17,8 +17,9 @@ package v0_6_exp
 import (
 	baseutil "github.com/coreos/butane/base/util"
 	"github.com/coreos/butane/config/common"
-
+	"github.com/coreos/ignition/v2/config/shared/errors"
 	"github.com/coreos/ignition/v2/config/util"
+	exp "github.com/coreos/ignition/v2/config/v3_5_experimental"
 	"github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
@@ -26,23 +27,64 @@ import (
 func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
 	var field string
 	sources := 0
+	var config string
+	var butaneReport report.Report
 	if rs.Local != nil {
 		sources++
 		field = "local"
+		config = *rs.Local
 	}
 	if rs.Inline != nil {
 		sources++
 		field = "inline"
+		config = *rs.Inline
 	}
 	if rs.Source != nil {
 		sources++
 		field = "source"
+		config = *rs.Source
 	}
 	if sources > 1 {
 		r.AddOnError(c.Append(field), common.ErrTooManyResourceSources)
 	}
+	if field == "local" || field == "inline" {
+		_, report, err := exp.Parse([]byte(config))
+		if len(report.Entries) > 0 {
+			butaneReport = ConvertToButaneReport(report, field)
+			r.Merge(butaneReport)
+		}
+		if err != nil {
+			r.AddOnError(c.Append(field), errors.ErrUnknownVersion)
+		}
+	}
 	return
 }
+
+func ConvertToButaneReport(ignitionReport report.Report, field string) report.Report {
+	var butaneRep report.Report
+	for _, entry := range ignitionReport.Entries {
+		
+		adjustedPath := []interface{}{field}
+		adjustedPath = append(adjustedPath, entry.Context.Path...)
+
+
+		butaneEntry := report.Entry{
+			Kind:    entry.Kind,
+			Message: entry.Message,
+			Context: path.ContextPath{
+				Path: adjustedPath,      // convert ignition path to butane path
+				Tag:  entry.Context.Tag,
+			},
+			Marker: entry.Marker,
+		}
+		butaneRep.Entries = append(butaneRep.Entries, butaneEntry)
+	}
+	return butaneRep
+}
+
+// func TranslatePath() {
+// path translating logic // TODO
+// }
 
 func (fs Filesystem) Validate(c path.ContextPath) (r report.Report) {
 	if !util.IsTrue(fs.WithMountUnit) {
--- a/base/v0_6_exp/validate_test.go
+++ b/base/v0_6_exp/validate_test.go
@@ -52,7 +52,7 @@ func TestValidateResource(t *testing.T) {
 		// inline specified
 		{
 			Resource{
-				Inline:      util.StrToPtr("hello"),
+				Inline:      util.StrToPtr("{\"ignition\": {\"version\": \"3.5.0\"}}"),
 				Compression: util.StrToPtr("gzip"),
 				Verification: Verification{
 					Hash: util.StrToPtr("this isn't validated"),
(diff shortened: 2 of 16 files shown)


bgilbert:
The file extension doesn't matter.  There's no requirement that Ignition configs have an `.ign` extension.  And, in the case of the `inline` directive, there isn't any file extension; the file contents are included directly in the parent config.

As #275 says, what we should do instead is validate the contents of the specified config using Ignition config validation.

Adam0Brien:
> As #275 says, what we should do instead is validate the contents of the specified config using Ignition config validation.

by ignition config validation do you mean  https://github.com/coreos/butane/blob/main/vendor/github.com/coreos/ignition/v2/config/v3_5_experimental/types/resource.go#L34 ?

bgilbert:
That's a validator that's invoked by the validation process.  We need to _run_ the validation process.

But thinking about it some more, we need to do more than just validation.  The embedded Ignition config is in the form of text, not Go structs, so we also need to parse it.  Ignition's config parsing function in `config` handles both parsing and validation, so we can just call that, and merge its report into ours.

A few notes:

- The parse function sometimes returns an `error` alongside an _empty_ report.  We should probably `AddOnError` the returned error to our validation report in addition to merging Ignition's report.

- Merging Ignition's report is a bit tricky.  We'd need to replace all the path fields with the path to the Butane field holding the config, which will be something like `$.ignition.config.merge.0.inline`.  That loses information but I'm not sure there's a good way around it.  We could try replacing the Ignition error in the log entry with a new one that includes the original path and the original message.

- The user is allowed to use older versions of Butane that don't understand the current Ignition spec, and is allowed to embed Ignition configs with versions newer than Butane understands.  We should have special handling for that case: https://github.com/coreos/butane/issues/275#issuecomment-1620636053

- This validation isn't FCOS-specific; it applies to all variants.  So, it should be in `base`.  And since we're only going to be producing errors for configs that were already invalid (but would be rejected when the machine boots rather than when Butane runs) we should backport it to stable `base` specs.

base/v0_6_exp/validate.go
```diff
@@ -29,10 +30,24 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
...
+			r.AddOnError(c.Append("ignition", "config", "merge", "local"), common.ErrEmptyReport)
+		}
 	}
 	if rs.Inline != nil {
 		sources++
 		field = "inline"
+		config := *rs.Inline
+		_, report, err := exp.Parse([]byte(config))
+		if err != nil {
+			r.Merge(report)
```
prestist, line 47: We cannot just merge this, the path fields in the report are ign specific. We need to take the report data and create a mapper function which translates the data into butane's paths

base/v0_6_exp/validate.go
```diff
@@ -29,10 +30,24 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
 	if rs.Local != nil {
 		sources++
 		field = "local"
+		config := *rs.Local
+		_, report, err := exp.Parse([]byte(config))
+		if err != nil {
+			r.Merge(report)
```
prestist, line 36: We cannot just merge this, the path fields in the report are ign specific. We need to take the report data and create a mapper function which translates the data into butane's paths

base/v0_6_exp/validate.go
```diff
@@ -29,10 +30,24 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
 	if rs.Local != nil {
```
prestist, line 30: This might be able to be compressed into a switch to reduce code duplication.

base/v0_6_exp/validate.go
```diff
@@ -29,10 +30,24 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
...
 		sources++
 		field = "local"
+		config := *rs.Local
+		_, report, err := exp.Parse([]byte(config))
+		if err != nil {
+			r.Merge(report)
+		} else {
+			r.AddOnError(c.Append("ignition", "config", "merge", "local"), common.ErrEmptyReport)
```
prestist, line 38: So why is it an error when there is no error reported from the ign parse function?

base/v0_6_exp/validate.go
```diff
@@ -41,7 +56,8 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
 	if sources > 1 {
 		r.AddOnError(c.Append(field), common.ErrTooManyResourceSources)
 	}
-	return
+
+	return r
```
prestist, line 60: I am pretty sure we dont need to return r explicitly here, we can keep it how it was.

base/v0_6_exp/validate.go
prestist: Do you mean if the report is empty then we error? if so the conditional logic here is wrong.

prestist (commented):
Hey, thank you for working on this I just took a first pass at this. 

I found some surface level questions for you. 

I am planning on taking another deeper pass so see why your CI is failing.

base/v0_6_exp/validate.go
Adam0Brien: no problem, I've made a start at the mapper function, I'm not too familiar with what exactly needs to be mapped but will look into it!

travier:
Let's split the vendor update into its own commit to make review easier.

base/v0_6_exp/validate.go
```diff
@@ -41,9 +44,39 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
...
+		}
+	}
+
 	return
 }
 
+func MapIgnitionReportToButane(ignitionReport report.Report) report.Report {
+	var butaneRep report.Report
+	for _, entry := range ignitionReport.Entries {
+		butaneEntry := report.Entry{
```
prestist, line 69: Im not sure this is everything we need.
Looping through the entries and expanding them into a new report object with those entries is good.
The issue comes with the the entry.Context path the paths are different between the ignition package and the butane package. this is mostly because we are converting from json to yaml butane is human readable so their names change.

internal to butane it looks like we already have a similar issue, we have to translate to json from yaml the direction is the inverse of what we need now though.  

It has a good test to explain this here => https://github.com/coreos/butane/blob/0244d31c627890b9af64b82e2a2224ac39bbb4d4/base/v0_6_exp/translate_test.go#L573C12-L573C12

prestist (commented):
Its getting really close, keep it up!

base/v0_6_exp/validate.go
```diff
@@ -41,9 +44,39 @@ func (rs Resource) Validate(c path.ContextPath) (r report.Report) {
 	if sources > 1 {
 		r.AddOnError(c.Append(field), common.ErrTooManyResourceSources)
 	}
+	if field == "local" || field == "inline" {
+		// check if it's an ignition config
+		if len(config) > 0 && config[0] == '{' {
```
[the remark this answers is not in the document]
Adam0Brien: I was thinking that the config variable was in ignition format so i was thinking to see if its valid t check the first character which is usually '{'. I will drop this

Issue #477, opened 2023-07-03
Import SElinux module


quentin9696:
# Feature Request #

## Environment ##

I run on AWS but I don't think it matter

## Desired Feature ##

Be able to import custom SELinux module with ignition. This is usually done by running the command `semodule -i /path/to/module`. Files can be compile module or `cil` modules

## Other Information ##

This must be done after the filesystem is loaded. 

Some OS comes with SElinux enabled by default, I think it would be nice to have more feature related to selinux

quentin9696:
Thanks @bgilbert

yasminvalim:
Hi! Butane is being merged into the Ignition repository, and the Butane repo will be archived soon.

We opened a new issue to continue this discussion in Ignition: https://github.com/coreos/ignition/issues/2267

Please use that issue for any follow-ups. Thank you!

PR #486, opened 2023-08-28
Import SElinux module #477

yasminvalim:
#### Import SElinux module 

We want to add sugar to butane which will allow users  import costum SElinux module. 

---

#### Related Issues

#477


travier:
Just a small note for your name in the Git config:
```
From: YASMIN VALIM <ydesouza@redhat.com>
```
You probably want `Yasmin Valim <ydesouza@redhat.com>` instead.

You'll have to modify it in you `~/.gitconfig` and then "reset" the author for all your commits here with git rebase or git commit --amend.

yasminvalim:
I'm now trying to test manually, and while reading the documentation, I had a question:
How can I be sure that my code is running locally?

I understand that I can build it with custom files that I created and it will be translated from BU to Ignition.
But I'm not sure how can I run the code I made instead of the butane version I installed and the latest image.

I think I am missing something.

prestist:
> I'm now trying to test manually, and while reading the documentation, I had a question: How can I be sure that my code is running locally?
> 
> I understand that I can build it with custom files that I created and it will be translated from BU to Ignition. But I'm not sure how can I run the code I made instead of the butane version I installed and the latest image.
> 
Noooo worries there are a lot of layers here.

So, to test it locally we have to exercise your code which lives in butane, then we want to test that our expectation of what is required for the system to effectively implement the sugar is correct which is done by exercising existing code in ignition through first booting a fcos image. 

The simplest step is directly exercising your code by creating a butane config which references the sugar you added, and using your locally built butane to create an .ign file.


- build your butane, and then `./butane.bin  --pretty --strict example.bu > example.ign`
Note you will want a way to get into your running fcos image so you can verify things in the OS after its running.
you can read a guide about auto login [here](https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-autologin/).
   - so to summarize your .bu will be the bu yaml in the example + your sugar.

Now to exercise your assumptions for enabling a semodule you would then need to first boot a fcos vm with your ignition config. The simpliest way to do this is by using cosa

1. You can either download a qcow image from [here](https://fedoraproject.org/coreos/download/?stream=stable) or build your own using cosa int, fetch, build

2. Then you can pass your ign file you built earlier using a command like so `$ cosa kola qemuexec --ignition ./ignition.json --ignition-direct --qemu-image ./yourImage.qcow2
`
3. If the ign config is understood and functional, you should be greeted by the console (with auto login) then you can do your probing of the os to see if the selinux module is enabled (as your sugar indicates)

travier:
You can find an small example SELinux module (CIL) in https://github.com/coreos/fedora-coreos-tracker/issues/1447#issuecomment-1528146989

yasminvalim:
Hey there,

I'm sharing my current approach for tackling this issue. I did have a few other ideas, but this one seems to be the most on point. I'm planning to commit the code I've been working on in the next few days, even though it might not be 100% ready, and it could potentially break my CI setup.

Here's the plan I've got in mind. If you've got any cool ideas or questions that could help me sharpen things up, please don't hold back. 

- Fields with content of the SElinux file and  the file path
- Create a filesystem (file: translate.go) (do I need it?)
- Create the file with that path and the content (file: translate.go)
- Kinda copy and paste the file 
- Execute the command “_semodule -i_” + path (file: translate.go)
- Translate butane to ignition  (file: translate.go)
- Validate some rules in validate.go:
    -  If there is a path, the content is required
    - If there is a content, the path is required
    - Any more rules required?
- Unit tests

This current approach would look like that in a butane file:

```
variant: fcos
version: 1.6.0-experimental
selinux_module:
  - contents:
      - |
        # setenforce 1
        # cat permissive-keyutils.cil
        (type keyutils_request_t)
        (typepermissive keyutils_request_t)
        # semodule -i permissive-keyutils.cil
  - path: /path/to/selinux/custom/file
```

Your input would be awesome and thanks for your attention.

travier:
The example Butane config should likely only include the content of the SELinux module:

```
variant: fcos
version: 1.6.0-experimental
selinux:
  module:
    - name: "permissive-keyutils"
    - contents: |
        (type keyutils_request_t)
        (typepermissive keyutils_request_t)
```

Then Butane would generate an Ignition config that:
- writes the content of this module to a file in `/etc/selinux` (not sure where exactly would be best)
- adds a systemd unit that loads this module on first boot using `semodule -i ...`.

prestist:
Your approach is sound. It might make sense to create tests before making the functions. In this case in particular could be well suited for TDD, and just create a test like [TestTranslateBootDevice](https://github.com/coreos/butane/blob/4996e45b5d67c8ef3b59e9c54db83f07417294bf/config/fcos/v1_6_exp/translate_test.go#L37C6-L37C29) which would fail until you create the translate code to build the sugar. 

Since you have a sound understanding of the input and the expected output. And as long as your spec changes are there you should be able to run the test with no compile errors.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +375,68 @@ func buildGrubConfig(gb Grub) string {
...
+			break
+		}
+
+		if hasValidModule {
+			rendered.Storage.Filesystems = append(rendered.Storage.Filesystems,
+				types.Filesystem{
+					Device: "/dev/disk/by-label/boot",
+					Format: util.StrToPtr("ext4"),
+					Path:   util.StrToPtr("/boot"),
+				})
```
prestist, line 401: So we only want this to happen once; but now it will happen for every valid module ? Im pretty sure we only want one filesystem.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +375,68 @@ func buildGrubConfig(gb Grub) string {
...
+
+			src, compression, err := baseutil.MakeDataURL([]byte(module.Content), nil, !options.NoResourceAutoCompression)
+			if err != nil {
+				r.AddOnError(yamlPath, err)
+				return rendered, ts, r
+			}
+
+			if module.Name != "" {
```
prestist, line 409: If we have already validated this I dont think we need to check again. maybe we could make a list or array of valid selinux modules when we validate them?

config/fcos/v1_6_exp/translate.go
```diff
@@ -382,53 +382,60 @@ func (c Config) handleSelinux(options common.TranslateOptions) (types.Config, tr
...
+	hasValidModule := false
 
-	// this should happen as many times as there are modules
 	for _, module := range c.Selinux.Module {
-
-		src, compression, err := baseutil.MakeDataURL([]byte(module.Content), nil, !options.NoResourceAutoCompression)
-		if err != nil {
-			r.AddOnError(yamlPath, err)
-			return rendered, ts, r
+		if module.Name != "" && module.Content == "" || module.Content != "" && module.Name == "" || module.Name == "" && module.Content == "" {
```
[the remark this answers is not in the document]
yasminvalim: That's right. Thank you!

config/fcos/v1_6_exp/translate.go
yasminvalim: You are correct, I removed that part of creating the file system from inside the for loop. I think makes more sense now.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +375,75 @@ func buildGrubConfig(gb Grub) string {
...
+
+	hasValidModule := false
+
+	for _, module := range c.Selinux.Module {
+		if isValidModule(module) {
+			hasValidModule = true
+			rendered = processModule(rendered, module, options, ts, r)
+			break
+		} else {
+			r.AddOnWarn(path.New("yaml", "selinux", "module"), common.ErrFieldInvalid)
```
prestist, line 392: you need to have the index here see other comment.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +375,75 @@ func buildGrubConfig(gb Grub) string {
...
+func (c Config) handleSelinux(options common.TranslateOptions) (types.Config, translate.TranslationSet, report.Report) {
+	rendered := types.Config{}
+	ts := translate.NewTranslationSet("yaml", "json")
+	var r report.Report
+
+	hasValidModule := false
+
+	for _, module := range c.Selinux.Module {
+		if isValidModule(module) {
+			hasValidModule = true
```
prestist, line 388: we can get rid of this. as we can look at our rendered types to see if there was a valid module

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +375,75 @@ func buildGrubConfig(gb Grub) string {
...
+			rendered = processModule(rendered, module, options, ts, r)
+			break
+		} else {
+			r.AddOnWarn(path.New("yaml", "selinux", "module"), common.ErrFieldInvalid)
+		}
+	}
+
+	if hasValidModule {
```
prestist, line 396: we should be able to say somthing like if leng(rendered.Storage.Files) > 0

yasminvalim:
Hey folks, I am still having problem with unit tests. Currently, I am having this error and those are the logs.
This error applies to each line in _translations._ 

```
--- FAIL: TestTranslateSelinux (0.00s)
    --- FAIL: TestTranslateSelinux/translate_0 (0.00s)
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:45: 
            	Error Trace:	/home/ydesouza/butane/base/util/test.go:45
            	            				/home/ydesouza/butane/config/fcos/v1_6_exp/translate_test.go:1717
            	Error:      	Not equal: 
            	            	expected: translate.Translation{From:path.ContextPath{Path:[]interface {}{"selinux", "module", 0}, Tag:"yaml"}, To:path.ContextPath{Path:[]interface {}{"storage"}, Tag:"json"}}
            	            	actual  : translate.Translation{From:path.ContextPath{Path:[]interface {}{"selinux", "module"}, Tag:"yaml"}, To:path.ContextPath{Path:[]interface {}{"storage"}, Tag:"json"}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,6 +2,5 @@
            	            	  From: (path.ContextPath) {
            	            	-  Path: ([]interface {}) (len=3) {
            	            	+  Path: ([]interface {}) (len=2) {
            	            	    (string) (len=7) "selinux",
            	            	-   (string) (len=6) "module",
            	            	-   (int) 0
            	            	+   (string) (len=6) "module"
            	            	   },
            	Test:       	TestTranslateSelinux/translate_0
            	Messages:   	non-identity translation with unexpected From
```

What I am missing here? 
Thanks! :smile_cat:

config/common/errors.go
```diff
@@ -93,6 +93,11 @@ var (
 
 	// Kernel arguments
 	ErrGeneralKernelArgumentSupport = errors.New("kernel argument customization is not supported in this spec version")
+
+	// Selinux Module
+	ErrContentInvalid = errors.New("Content is empty, please provide content.")
+	ErrNameInvalid    = errors.New("Name is empty, please provide a valid name.")
+	ErrFieldInvalid   = errors.New("Please, provide valid information.")
```
jlebon, line 100: If these are all specific to SELinux, we should probably prefix them with e.g. `ErrSelinux`.

`ErrFieldInvalid` seems redundant.

For the error messages, let's use the same wording as other errors, e.g. "field ... is required".

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +374,79 @@ func buildGrubConfig(gb Grub) string {
...
 }
+
+func (c Config) handleSelinux(options common.TranslateOptions) (types.Config, translate.TranslationSet, report.Report) {
+	rendered := types.Config{}
+	ts := translate.NewTranslationSet("yaml", "json")
+	var r report.Report
+
+	for i, module := range c.Selinux.Module {
+		yamlPath := path.New("yaml", "selinux", "module", i)
+		if module.Name != "" && module.Content != "" {
```
jlebon, line 385: Shouldn't this already be checked by validation?

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +374,79 @@ func buildGrubConfig(gb Grub) string {
...
+		newFilesystem := types.Filesystem{
+			Device: "/dev/disk/by-label/boot",
+			Format: util.StrToPtr("ext4"),
+			Path:   util.StrToPtr("/boot"),
+		}
+
+		rendered.Storage.Filesystems = append(rendered.Storage.Filesystems, newFilesystem)
+		ts.AddFromCommonSource(path.New("yaml", "selinux", "module", 0), path.New("json", "storage"), newFilesystem)
+
+	}
```
jlebon, line 402: This sugar shouldn't need to add filesystem things.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +374,79 @@ func buildGrubConfig(gb Grub) string {
...
+			},
+		},
+	}
+
+	filePath := path.New("json", "storage", "files", len(rendered.Storage.Files), newFile)
+	rendered.Storage.Files = append(rendered.Storage.Files, newFile)
+	ts.AddFromCommonSource(yamlPath, filePath, newFile)
+
+	// Create systemd unit to import module
+	cmdToExecute := "/usr/sbin/semodule -i" + modulePath
```
jlebon, line 435:
```suggestion
	cmdToExecute := "/usr/sbin/semodule -i " + modulePath
```

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +374,79 @@ func buildGrubConfig(gb Grub) string {
...
+	newUnit := types.Unit{
+		Name:    "selinux-import-module@.service",
```
jlebon, line 437: Is there a reason we're creating a template unit here instead of just directly creating a unit like `selinux-import-module-$name.service`?

config/fcos/v1_6_exp/translate_test.go
```diff
@@ -1637,3 +1637,104 @@ func TestTranslateGrub(t *testing.T) {
...
+func TestTranslateSelinux(t *testing.T) {
+	cmdToExecute := "/usr/sbin/semodule -i" + "/etc/selinux/targeted/modules/active/extra/some_name.cil"
+	translations := []translate.Translation{
+		{From: path.New("yaml", "version"), To: path.New("json", "ignition", "version")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "filesystems")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "filesystems", 0)},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "filesystems", 0, "path")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "filesystems", 0, "device")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "filesystems", 0, "format")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage")},
```
jlebon, line 1650: Related to the other comment, we should be able to remove all of these.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +374,79 @@ func buildGrubConfig(gb Grub) string {
...
+		Enabled: util.BoolToPtr(true),
+		Dropins: []types.Dropin{
+			{
+				Name:     module.Name + ".conf",
+				Contents: util.StrToPtr(" [Unit]\n" + "Description=Import SELinux module\n" + "[Service]\n" + "Type=oneshot\n" + "RemainAfterExit=no\n" + "ExecStart=" + cmdToExecute + "\n" + "[Install]\n" + "WantedBy=multi-user.target\n"),
```
jlebon, line 442: Should include the module name in the description.
Also, shouldn't `RemainAfterExit` be `yes`?

config/fcos/v1_6_exp/validate.go
```diff
@@ -77,3 +77,14 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
...
 }
+
+func (m Module) Validate(c path.ContextPath) (r report.Report) {
+	if m.Name != "" && m.Content == "" {
+		r.AddOnError(c.Append("content"), common.ErrContentInvalid)
+	}
+
+	if m.Content != "" && m.Name == "" {
+		r.AddOnError(c.Append("name"), common.ErrNameInvalid)
+	}
```
jlebon, line 88: I don't follow the logic here. Why not just check both independently, possibly adding both errors if both fields are empty?

jlebon (commented):
I'm not an SME on this codebase and don't have the cycles to be one, unfortunately. :( I did a semi-shallow review of what I could piece together.

config/fcos/v1_6_exp/validate.go
yasminvalim: I was trying not to generate any errors if both were empty in case the user doesn't fill in this. 

But your idea makes more sense. I'll change that. 

Thank you!

config/fcos/v1_6_exp/translate.go
yasminvalim: I agree. Thanks!

config/fcos/v1_6_exp/translate_test.go
yasminvalim: All the filesystem stuff, right?

config/fcos/v1_6_exp/translate.go
yasminvalim: Not really. How can I create it directly?

config/fcos/v1_6_exp/translate.go
yasminvalim: Just curious to know how it works:
Why we don't need this? 
Should I remove all the filesystem part?

Thanks!

config/fcos/v1_6_exp/translate.go
yasminvalim: Not sure. Discussed with @steven, and validations don't seem to impact this part of the code and we should have some rules here as well.

I can't explain it very well; perhaps Steven can provide some insights about it.

config/common/errors.go
yasminvalim: Nice! I will change it! Thanks.

config/fcos/v1_6_exp/translate.go
jlebon: Hmm, odd. If I compare to the GRUB code, here's what I see.

We validate e.g. the name here:

https://github.com/coreos/butane/blob/310a1f1bd2725e01f0805e637c4cb39bc36e5763/config/fcos/v1_5/validate.go#L45C1-L47

And here in `translate.go` we just add the name without any validation:

https://github.com/coreos/butane/blob/310a1f1bd2725e01f0805e637c4cb39bc36e5763/config/fcos/v1_5/translate.go#L362C1-L363

Note also the comment there which makes it seem like we run through validation first.

jlebon: If I look at the high-level `TranslateBytes()` function, it does seem like we validate before doing the translation: https://github.com/coreos/butane/blob/310a1f1bd2725e01f0805e637c4cb39bc36e5763/config/util/util.go#L111

config/fcos/v1_6_exp/translate.go
jlebon: In the GRUB case, we need to add a file in the boot filesystem (`user.cfg`) which is why we need Ignition to mount it so we can write to it. In our case, all the logic happens in the rootfs, so there's no need to edit the bootfs.

config/fcos/v1_6_exp/translate.go
jlebon: It's very similar to what you have now but instead of defining a dropin, we define the contents of the systemd unit itself.

config/fcos/v1_6_exp/translate_test.go
jlebon: Yup!

config/fcos/v1_6_exp/translate.go
prestist: @yasminvalim yeah @jlebon is certainly correct, sorry about the miss-info. I completely spaced on this. Thank you for the links!

config/fcos/v1_6_exp/translate.go
prestist: Ah that makes sense; yeah we were using the GRUB case as our backing for this and incorrectly assumed the filesystem was a requirement.

config/fcos/v1_6_exp/translate.go
yasminvalim: Thanks for explain! I changed it!

yasminvalim:
I've made few changes based on the review and debugged the code. However, I'm still uncertain about what's causing the issue in either the code or the test itself. Below is the current error. 
```
Running tool: /usr/bin/go test -timeout 30s -run ^TestTranslateSelinux$ github.com/coreos/butane/config/fcos/v1_6_exp

--- FAIL: TestTranslateSelinux (0.00s)
    --- FAIL: TestTranslateSelinux/translate_0 (0.00s)
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module → $.storage
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0.path
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0.append
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0.append.0
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0.append.0.source
        /home/ydesouza/butane/config/fcos/v1_6_exp/test.go:47: missing non-identity translation $.selinux.module.0 → $.storage.files.0.append.0.compression
        /home/ydesouza/butane/config/fcos/v1_6_exp/translate_test.go:1737: 
            	Error Trace:	/home/ydesouza/butane/config/fcos/v1_6_exp/translate_test.go:1737
            	Error:      	Received unexpected error:
            	            	Missing paths in TranslationSet:
            	            	$.storage.files.0.path
            	            	$.storage.files.0.append.0.compression
            	            	$.storage.files.0.append.0.source
            	            	$.storage.files.0.append.0
            	            	$.storage.files.0.append
            	            	$.storage.files.0
            	            	$.storage.files
            	            	$.storage
            	            	$.systemd.units
            	            	$.systemd
            	Test:       	TestTranslateSelinux/translate_0
            	Messages:   	incomplete TranslationSet coverage
FAIL
FAIL	github.com/coreos/butane/config/fcos/v1_6_exp	0.003s
FAIL
```

Please share any insights if you have them. :smile_cat:

config/fcos/v1_6_exp/schema.go
```diff
@@ -22,6 +22,8 @@ type Config struct {
 	base.Config `yaml:",inline"`
 	BootDevice  BootDevice `yaml:"boot_device"`
 	Grub        Grub       `yaml:"grub"`
+	Selinux     Selinux    `yaml:"selinux"`
+	Systemd     Systemd    `yaml:"systemd"`
```
jlebon, line 26: Hmm, why are we overriding the `systemd` key? I think we only need this if we need to change something in how the toplevel `systemd` Butane key gets translated, but that's not the case here (our `selinux` sugar leverages the `systemd` field in the Ignition config, but shouldn't affect the Butane `systemd` schema).

config/fcos/v1_6_exp/schema.go
```diff
@@ -49,3 +51,28 @@ type GrubUser struct {
...
+	Dropins  []Dropin `yaml:"dropins"`
+	Enabled  *bool    `yaml:"enabled"`
+	Mask     *bool    `yaml:"mask"`
+	Name     string   `yaml:"name"`
+}
+
+type Dropin struct {
+	Contents *string `yaml:"contents"`
+	Name     string  `yaml:"name"`
+}
```
jlebon, line 70: So then, we would delete this.

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +378,60 @@ func buildGrubConfig(gb Grub) string {
...
+				"Description=Import SELinux module\n" +
+				"[Service]\n" +
+				"Type=oneshot\n" +
+				"RemainAfterExit=yes\n" +
+				"ExecStart=" + cmdToExecute + "\n" +
+				"[Install]\n" +
+				"WantedBy=multi-user.target\n"),
+		Enabled: util.BoolToPtr(true),
+	})
+	ts.AddFromCommonSource(yamlPath, path.New("json", "systemd"), rendered.Systemd)
```
prestist, line 434: So while this certainly passes the validate, I am having difficulty understanding why this should not be 

 `ts.AddFromCommonSource(yamlPath, path.New("json", "systemd","units",len(rendred.systemd.units)), rendered.Systemd.units)`

config/fcos/v1_6_exp/translate.go
```diff
@@ -367,3 +378,60 @@ func buildGrubConfig(gb Grub) string {
...
+		FileEmbedded1: types.FileEmbedded1{
+			Append: []types.Resource{
+				{
+					Source:      util.StrToPtr(src),
+					Compression: compression,
+				},
+			},
+		},
+	})
+	ts.AddFromCommonSource(yamlPath, path.New("json", "storage"), rendered.Storage)
```
prestist, line 416: So while this certainly passes the validate, I am having difficulty understanding why this should not be
 
 `ts.AddFromCommonSource(yamlPath, path.New("json", "storage","files",len(rendred.storage.files)), rendered.storage.files)`

prestist (commented):
We did a debug session on this, and finally got to a workable point, but we remain puzzled on these two things @jlebon

prestist:
@yasminvalim the go 1.21 test should be fixed by https://github.com/coreos/butane/pull/505.
<details>
<summary>steps to update branch</summary>
assuming origin is your fork and upstream is 

```
git fetch upstream
git merge upstream/main 
git push -f
```
</details>

prestist:
> @yasminvalim the go 1.21 test should be fixed by #505.
> steps to update branch
> assuming origin is your fork and upstream is

Sorry I led you astray, the merge commit is nooooo good, the best thing to do is likely to undo your merge and just rebase

git revert -m 1 <commit-hash>
git fetch upstream
git rebase upstream/main

that way your fast-forwarded rather then merged into.

yasminvalim:
> > @yasminvalim the go 1.21 test should be fixed by #505.
> > steps to update branch
> > assuming origin is your fork and upstream is
> 
> Sorry I led you astray, the merge commit is nooooo good, the best thing to do is likely to undo your merge and just rebase
> 
> git revert -m 1 [a419de4](https://github.com/coreos/butane/commit/a419de4df7286bb307f87df55763db153fa418ef) git fetch upstream git rebase upstream/main
> 
> that way your fast-forwarded rather then merged into.

I think it's  fine now, but not entirely sure because it's showing all the files updated in the old merge. Can you take a look, please?

Also,  can I use  `rebase` as a pattern when I need to update a branch?

prestist:
@yasminvalim I am a bit confused because this https://github.com/coreos/butane/pull/486/commits/36f71f1e0d0660796212b1ab0dc50b1b91e3379c should not be a thing. after reverting your merge it should have put you in a place with only your other commit. 

Once you revert the merge you can update your branch with the rebase pattern instead.

config/fcos/v1_6_exp/translate.go
```diff
@@ -376,3 +387,60 @@ func buildGrubConfig(gb Grub) string {
...
 	return "# Generated by Butane\n\n" + superUserCmd + strings.Join(cmds, "\n") + "\n"
 }
+
+func (c Config) handleSelinux(options common.TranslateOptions) (types.Config, translate.TranslationSet, report.Report) {
+	rendered := types.Config{}
+	ts := translate.NewTranslationSet("yaml", "json")
+	var r report.Report
+
+	for _, module := range c.Selinux.Module {
+		rendered = processModule(rendered, module, options, ts, r, path.New("yaml", "selinux", "module"))
```
jlebon, line 397: We should add the index to the path here so that e.g. module 2 is associated with the right output files and systemd units.

config/fcos/v1_6_exp/translate.go
jlebon: I think I understand this better now.

So I do think this is correct. It's saying that everything under and including the `.storage` key is associated with that module (after adding the index like I mentioned above, it'd be `.selinux.module.0 -> .storage` and `-> .storage.files` and so on). Essentially, we're not just adding a file and a unit because of this module, we have to add all the keys leading up to it. All keys in the output need a source translation associated, so e.g. `.storage` *needs* to be associated to *something* on the YAML side.

config/fcos/v1_6_exp/translate.go
jlebon: And I think this is correct too.

config/fcos/v1_6_exp/translate_test.go
```diff
@@ -1637,3 +1637,97 @@ func TestTranslateGrub(t *testing.T) {
...
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "files", 0, "append")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "files", 0, "append", 0)},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "files", 0, "append", 0, "source")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "storage", "files", 0, "append", 0, "compression")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd", "units", 0, "name")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd", "units", 0, "contents")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd", "units", 0, "enabled")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd", "units", 0)},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd", "units")},
+		{From: path.New("yaml", "selinux", "module"), To: path.New("json", "systemd")},
```
jlebon, line 1658: With the added index, you'd add 0 to all the From paths here.

config/fcos/v1_6_exp/translate.go
```diff
@@ -376,3 +387,60 @@ func buildGrubConfig(gb Grub) string {
...
+	ts.AddFromCommonSource(yamlPath, path.New("json", "storage"), rendered.Storage)
+
+	// Create systemd unit to import module
+	cmdToExecute := "/usr/sbin/semodule -i" + modulePath
+
+	rendered.Systemd.Units = append(rendered.Systemd.Units, types.Unit{
+		Name: module.Name + ".conf",
+		Contents: util.StrToPtr(
+			"[Unit]\n" +
+				"Description=Import SELinux module\n" +
```
jlebon, line 434: This description should be templated.

config/fcos/v1_6_exp/translate.go
```diff
@@ -376,3 +387,60 @@ func buildGrubConfig(gb Grub) string {
...
+		},
+	})
+	ts.AddFromCommonSource(yamlPath, path.New("json", "storage"), rendered.Storage)
+
+	// Create systemd unit to import module
+	cmdToExecute := "/usr/sbin/semodule -i" + modulePath
+
+	rendered.Systemd.Units = append(rendered.Systemd.Units, types.Unit{
+		Name: module.Name + ".conf",
+		Contents: util.StrToPtr(
```
jlebon, line 432: I would add a comment here (in the generated systemd unit) like:

```
# Generated by Butane
```

config/fcos/v1_6_exp/validate.go
```diff
@@ -98,3 +98,20 @@ func (user GrubUser) Validate(c path.ContextPath) (r report.Report) {
 	}
 	return
 }
+
+func (m Module) Validate(c path.ContextPath) (r report.Report) {
+	if m.Name == "" && m.Content == "" {
+		r.AddOnError(c.Append("name"), common.ErrSelinuxContentNotSpecified)
+		r.AddOnError(c.Append("content"), common.ErrSelinuxContentNotSpecified)
+	} else {
```
jlebon, line 106: I don't think we need this.

config/fcos/v1_6_exp/validate_test.go
```diff
@@ -479,3 +479,49 @@ func TestValidateConfig(t *testing.T) {
...
+		errPath path.ContextPath
+	}{
+		{
+			// valid module
+			in: Module{
+				Content: "some content",
+				Name:    "some name",
+			},
+			out:     nil,
+			errPath: path.New("yaml"),
```
jlebon, line 496: Hmm, we shouldn't need this.

config/fcos/v1_6_exp/validate.go
yasminvalim: I removed both this implementation and test.

config/fcos/v1_6_exp/translate.go
yasminvalim: Done, I've added the comment. Thanks. :smile_cat:

config/fcos/v1_6_exp/translate.go
yasminvalim: I saw in another file how it's done and managed to reproduce. Thank you for letting me know :cat:

config/fcos/v1_6_exp/translate.go
yasminvalim: Added the index path in the implementation and in the 'From Path'  in test files. :+1:

config/fcos/v1_6_exp/schema.go
```diff
@@ -50,3 +51,12 @@ type GrubUser struct {
 	Name         string  `yaml:"name"`
 	PasswordHash *string `yaml:"password_hash"`
 }
+
+type Selinux struct {
+	Module []Module `yaml:"module"`
```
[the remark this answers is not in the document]
yasminvalim: Nice! I changed it and also altered content to plural since the other examples are also using plural.

config/fcos/v1_6_exp/schema.go
```diff
@@ -50,3 +51,12 @@ type GrubUser struct {
...
+}
+
+type Module struct {
+	Name    string `yaml:"name"`
+	Content string `yaml:"content"`
```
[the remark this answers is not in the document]
yasminvalim: I did it, but I'm not sure if it's good enough. 
I created a type 'resource'  in v1_6_exp/schema as in the other versions to include it as a type for 'contents'. 

The tests are working even with this change, so I think it's okay.

config/fcos/v1_6_exp/validate.go
yasminvalim: Thanks for your review :)

yasminvalim:
Since Butane is now merged with Ignition we created a new issue related to this for Ignition: https://github.com/coreos/ignition/issues/2267

Issue #488, opened 2023-09-06
Stabilize OCP 4.14 spec


travier:
# Bumping spec versions

## Bump distro version

- [x] Rename `config/distro/vD_exp` to `config/distro/vD` and update `package` statements. Update imports.
- [x] Drop `-experimental` from `init()` in `config/config.go`.
- [x] Drop `-experimental` from examples in `docs/`.
- [x] Copy `config/distro/vD` to `config/distro/vD+1_exp`.
- [x] Update `package` statements in `config/distro/vD+1_exp`. Bump its base dependency to `base/vB+1_exp`.
- [x] Import `config/vD+1_exp` in `config/config.go` and add `distro` `C+1-experimental` to `init()`.

## Update docs

- [x] Update `internal/doc/main.go` to add the new stable spec and reference the new experimental spec in `generate()`.
- [x] Run `generate` to regenerate spec docs.
- [x] Update `docs/specs.md`.
- [x] Update `docs/upgrading-*.md` for the new spec version. Copy the relevant section from Ignition's `doc/migrating-configs.md`, convert the configs to Butane configs, convert field names to snake case, and update wording as needed. Add subsections for any new Butane-specific features.
- [x] Note the stabilization in `docs/release-notes.md`, following the format of previous stabilizations. Drop the `-exp` version suffix from any notes for the upcoming release.

travier:
We might want to revert https://github.com/coreos/butane/commit/4bd006e3f00eae89f9555e986b331fce8fdc2882 from https://github.com/coreos/butane/pull/456 as it does not work with the MCO in 4.14.

travier:
https://github.com/coreos/butane/issues/490

travier:
On top of the existing mapping that we have here between Butane and Ignition spec versions, it would be nice to get a doc update in https://github.com/openshift/machine-config-operator/tree/master/docs to add a mapping between Ignition specs and the branch in which they are supported / used as default internal spec in the MCO.

- https://github.com/openshift/machine-config-operator/blob/5b821a279c88fee1cc1886a6cf1ec774891a2258/pkg/controller/common/helpers.go#L609
- https://github.com/openshift/machine-config-operator/pull/3576
- https://github.com/openshift/machine-config-operator/pull/3814

PR #438, opened 2023-03-13
Stabilize openshift 4.13.0; create openshift 4.14.0-experimental

prestist:
Roll back to Ignition 3.2.0 spec for #312.

PR #491, opened 2023-09-14
Stabilize openshift 4.14.0; create openshift 4.15.0-experimental

travier:
See checklist in https://github.com/coreos/butane/issues/488.

---

openshift/v4_15_exp: Copy from 4.14_exp

---

openshift/v4_15_exp: Update spec to 4.15.0-experimental

---

openshift/v4.14: Stabilize 4.14.0 spec

---

openshift/v4.14: Use fcos 1.5 & ignition 3.4 specs

Stabilise openshift 4.14.0 spec on fcos 1.5 and ignition 3.4 stable
specs. The MCO is now capable of understanding Ignition 3.4.0 specs and
uses that by default.

See: https://github.com/openshift/machine-config-operator/pull/3576
See: https://github.com/openshift/machine-config-operator/pull/3814

---

openshift/v4.14: Remove GRUB password support

Remove all GRUB bootloader configuration sugar for openshift/v4.14
configs as the MCO does not know how to manage files in the read only
/boot partition yet. While we could use Ignition to write the initial
file, any update would fail which makes for a bad user experience.

This is revert of [1] for openshift/v4.14 only.

[1] https://github.com/coreos/butane/commit/4bd006e3f00eae89f9555e986b331fce8fdc2882
See: https://issues.redhat.com/browse/OCPSTRAT-831
See: https://issues.redhat.com/browse/OCPSTRAT-113

---

NEWS: Update for 4.14 stabilization

Fixes: https://github.com/coreos/butane/issues/488

---

References for this PR:
- https://github.com/coreos/butane/pull/438 previous stabilization PR


travier:
Currently trying to figure out if the MCO supports swap in 4.14 or not.

prestist (commented):
From what I see, everything looks correct. My only nit is the commit message for removing the Grub password. It would be helpful to try and document why in the commit message.

prestist:
> We likely don't support setting up swap yet in openshift (see https://kubernetes.io/blog/2023/08/24/swap-linux-beta/) so maybe we should remove those mentions from the docs.

Hmm, thats a bit sticky. Once this is stabilized I suspect updates continue to flow into open shift 4.14.0?
If so then is there not the likelihood that it will support swap?

travier:
I've update some docs elements and finalized the commit messages. This should be ready for a final review and then we can make a new release.

prestist (approved):
LGTM, thank you for doing this :)

PR #517, opened 2024-02-19
docs/release-notes: update for release

yasminvalim:
Update the release notes for Butane 0.20.0 as per the release checklist issue #512.

PR #514, opened 2024-02-16
Stabilise openshift 4.15 spec & add openshift 4.16-experimental spec

travier:
docs: Fix openshift v4.14 generated docs

Fixes: https://github.com/coreos/butane/commit/186b8e8a7e2183b470f3edee967ca3da12761153

---

openshift/v4.15: Stabilize 4.15.0 spec

---

openshift/v4.15: Use fcos 1.5 & ignition 3.4 specs

Stabilise openshift 4.14.0 spec on fcos 1.5 and ignition 3.4 stable
specs. fcos 1.6 & ignition 3.5 are still experimental.

---

internal/doc/butane: Fix whitespace

Fixes: 0d2fa0f spec/v1.6: Add support for boot_device sugar on s390x

---

openshift/v4.15: Remove GRUB password support

---

openshift/v4_16_exp: Add 4.16.0-experimental spec

--

Fixes: https://github.com/coreos/butane/issues/511
Based on https://github.com/coreos/butane/pull/513 for convenience.

PR #484, opened 2023-08-12
fcos: add s390x boot_device sugar

madhu-pillai:
Hi,
 
I've closed the earlier pr 483 due to some errors. Here is the new PR.

As per the GitHub issue [https://github.com/coreos/butane/issues/453#issuecomment-1655441505]. The PR contains following conditional changes exclusively for s390x.

1. Added layout specific to s390x .
- > s390x-zfcp  s390x-eckd and s390x-virt .

2.   Added boot_device.luks.device specifically for s390x-zfcp s390x-eckd and forbidden for other arch including s390x-virt.

3.  The configuration will generate only with boot_device.luks and it fails if mirror boot_device.mirror  configuration is specified for s390x-eckd and s390x-zfcp .

PR #518, opened 2024-02-20
docs: Move s390x boot_devices sugar to ocp 4.16.0-experimental

travier:
It depends on fcos 1.6.0-experimental

See: https://github.com/coreos/butane/pull/484
See: https://github.com/coreos/butane/pull/514
Fixes: https://github.com/coreos/butane/pull/517


prestist:
Butane has already tagged; and the release notes for v20 are locked in, I would recommend making this change in v21, with a fix*

jmarrero:
I see this is all for docs, does this affect the functionality that is available on the release? I am not sure if that yaml change makes the feature available/unavailable on a specific version.

prestist:
Actually yeah you're right. Thank you for the point.

travier:
Yes, this is all docs so we don't need another release.

PR #523, opened 2024-04-01
flatcar/v1_2_exp: Add clevis support

simoncampion:
This PR updates `flatcar/v1_2_exp` to reflect Flatcar can handle disk encryption with clevis, since [this commit](https://github.com/flatcar/scripts/pull/1560).

### DIFF
--- a/config/flatcar/v1_2_exp/translate.go
+++ b/config/flatcar/v1_2_exp/translate.go
@@ -22,15 +22,9 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
-var (
-	fieldFilters = cutil.NewFilters(types.Config{}, cutil.FilterMap{
-		"storage.luks.clevis": common.ErrClevisSupport,
-	})
-)
-
 // Return FieldFilters for this spec.
 func (c Config) FieldFilters() *cutil.FieldFilters {
-	return &fieldFilters
+	return nil
 }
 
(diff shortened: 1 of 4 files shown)


config/flatcar/v1_2_exp/translate.go
```diff
@@ -22,15 +22,9 @@ import (
...
-var (
-	fieldFilters = cutil.NewFilters(types.Config{}, cutil.FilterMap{
-		"storage.luks.clevis": common.ErrClevisSupport,
-	})
-)
-
 // Return FieldFilters for this spec.
 func (c Config) FieldFilters() *cutil.FieldFilters {
-	return &fieldFilters
+	return &cutil.FieldFilters{}
```
pothos, line 27:
```suggestion
	return nil
```
Thanks for the PR! The only thing I would see is that the other `translate.go` files use `nil` here

simoncampion: Changed it to `nil` here. (This also required a tiny change to the test logic, handling the case of `FieldFilters()` returning `nil`.)

prestist (commented):
Awesome, thank you for the update! 
The changes lgtm, but could you add a comment in the release notes please :)

simoncampion:
@prestist Thanks! I added an entry in the release notes and ran `./generate`.

prestist (approved):
Thank you for working on this. Overall LGTM!

PR #524, opened 2024-04-09
examples: correct partition number for var example

prestist:


travier:
I don't think this is correct. This is an example for creating a partition on a secondary disk, so there are not other partitions there.

prestist:
Ugh, `sdb` should had been a pointer, yeah your right. Good catch!

PR #530, opened 2024-05-14
Fixed 'append' to 'appends'

TeamLinux01:
With `append`, I was getting an error of:
```
Error translating config: yaml: unmarshal errors:
  line 81: cannot unmarshal !!map into []v0_5.Resource
```

Switching to `appends` fixes the issue.


TeamLinux01:
Sorry, I didn't notice I got a warning.
```
warning at $.storage.files.0.appends, line 72 col 7: Unused key appends
```

PR #540, opened 2024-06-12
Dockerfile: update to Fedora 40

aaradhak:
https://github.com/coreos/fedora-coreos-tracker/issues/1674

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM registry.fedoraproject.org/fedora:39 AS builder
+FROM registry.fedoraproject.org/fedora:40 AS builder
 RUN dnf install -y golang git-core
 RUN mkdir /butane
 COPY . /butane
 WORKDIR /butane
 RUN ./build_for_container
 
-FROM registry.fedoraproject.org/fedora-minimal:39
+FROM registry.fedoraproject.org/fedora-minimal:40
 COPY --from=builder /butane/bin/container/butane /usr/local/bin/butane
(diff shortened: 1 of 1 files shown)


aaradhak:
@jlebon I see that the release notes are already updated to Fedora 40 in this case - https://github.com/coreos/butane/blob/main/docs/release-notes.md

jlebon:
> @jlebon I see that the release notes are already updated to Fedora 40 in this case - [`main`/docs/release-notes.md](https://github.com/coreos/butane/blob/main/docs/release-notes.md?rgh-link-date=2024-06-12T21%3A09%3A27Z)

Not quite. That release note entry refers to the `butane` binary. The container image itself being based on Fedora 40 is a separate change. Though it doesn't look like historically we've release noted it.

That said, this is missing another instance in the same file.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:39 AS builder
+FROM registry.fedoraproject.org/fedora:40 AS builder
```
prestist, line 1: Do this but also on the next "FROM" line.

prestist:
@aaradhak  after you have updated both From's you can then add a label for "skip notes" as we dont generally doc the update of this.

Dockerfile
aaradhak: missed that one. Thanks for spotting that.

jlebon (approved):

PR #547, opened 2024-09-18
Stabilize openshift v4.17.0 spec

prestist:
Fixes: #545


docs/specs.md
```diff
@@ -25,6 +25,8 @@ We recommend that you always use the latest **stable** specification for your op
   - [v1.1.0](config-flatcar-v1_1.md)
   - [v1.0.0](config-flatcar-v1_0.md)
 - OpenShift (`openshift`)
+  - [v4.17.0](config-openshift-v4_17.md)
+  - [v4.16.0](config-openshift-v4_16.md)
```
travier, line 29: Ouch this was missing from the 4.16 PR. Good that we're fixing it here :)

docs/release-notes.md
```diff
@@ -94,7 +98,7 @@ key](https://getfedora.org/security/).
 ### Docs changes
 
 - Document consequence of setting `systemd.units.mask` to false
-- Document `grub` section _(openshift 4.15.0-exp)_
+- Document `grub` section _(openshift 4.18.0-exp)_
```
travier, line 101: Maybe I did this in the past, but let's try not changing past release notes.

We can add a note for the current release instead.

Someting like:
```
Support and documentation for `grub` section moved to OpenShift 4.18.0-experimental spec.
```

travier (approved):
Looks, great! Thanks. I only pushed a few nits.

docs/specs.md
prestist: yeaaah

docs/release-notes.md
prestist: Ah, ok, makes sense. will not do in the future; thank you for fixing the nits!

Issue #558, opened 2024-11-19
Bump `ignition` experimental to 3.6.0-experimental


travier:
# Bumping spec versions

This checklist describes bumping the Ignition spec version, `base` version, and distro versions. If your scenario is different, modify to taste.

## Stabilize Ignition spec version

- [x] Bump `go.mod` for new Ignition release and update vendor.
- [x] Update imports. Drop `-experimental` from Ignition spec versions in `*/translate_test.go`.

## Bump base version

- [x] Rename `base/vB_exp` to `base/vB` and update `package` statements. Update imports.
- [x] Copy `base/vB` to `base/vB+1_exp`.
- [x] Update `package` statements in `base/vB+1_exp`.

## Bump distro version

- [x] Rename `config/distro/vD_exp` to `config/distro/vD` and update `package` statements. Update imports.
- [x] Drop `-experimental` from `init()` in `config/config.go`.
- [x] Drop `-experimental` from examples in `docs/`.
- [x] Copy `config/distro/vD` to `config/distro/vD+1_exp`.
- [x] Update `package` statements in `config/distro/vD+1_exp`. Bump its base dependency to `base/vB+1_exp`.
- [x] Import `config/vD+1_exp` in `config/config.go` and add `distro` `C+1-experimental` to `init()`.

## Bump Ignition spec version

- [x] Bump Ignition types imports and rename `ToIgnI` and `TestToIgnI` functions in `base/vB+1_exp`. Bump Ignition spec versions in `base/vB+1_exp/translate_test.go`.
- [x] Bump Ignition types imports in `config/distro/vD+1_exp`. Update `ToIgnI` function names, `util` calls, and header comments to `ToIgnI+1`.

## Update docs

- [x] Update `internal/doc/main.go` to add the new stable spec and reference the new experimental spec in `generate()`.
- [x] Run `generate` to regenerate spec docs.
- [x] Update `docs/specs.md`.
- [ ] Update `docs/upgrading-*.md` for the new spec version. Copy the relevant section from Ignition's `doc/migrating-configs.md`, convert the configs to Butane configs, convert field names to snake case, and update wording as needed. Add subsections for any new Butane-specific features.
- [x] Note the stabilization in `docs/release-notes.md`, following the format of previous stabilizations. Drop the `-exp` version suffix from any notes for the upcoming release.

prestist:
Closed by #562

PR #562, opened 2024-11-22
Ignition spec bump

prestist:
https://github.com/coreos/butane/issues/558


travier:
Hum, it looks like you're trying to stabilize a lot of variants in this PR at the same time.

travier:
I had copied the entire template in the issue but we don't need to do it all.

We only need to bump the experimental versions.

travier:
So, I took another look at this and what I would do is to only bump ignition to 3.6-exp first in all the experimental config variants.

Then in https://github.com/coreos/butane/issues/559, we would stabilise the base, fcos and 4.18 config variants on 3.5.0 and keep all of the other variants on the new exp base & fcos ones.

prestist:
spoke with @travier ; dialing back changes to go into independent prs, https://github.com/coreos/butane/issues/550#issuecomment-2488967415 . You can see that this pr is simply an import of the latest ignition with its now stabilized 3.5, which resulted in some new docs, see second commit.

travier:
I considered splitting the vendor part from the changes in the first commit but that would have made it non-atomic / harder to bissect while making it easier to distinct which changes were manual vs the vendor code.

But overall it's the same result so let's go with that as it's bissectable.

travier (approved):

travier:
Hum, maybe we should move the spec that we are not going to stabilize directly to 3.6.0-exp in this one.

prestist:
> Hum, maybe we should move the specs that we are not going to stabilize directly to 3.6.0-exp in this one.

Thats fair, I was thinking I would do that when adding the new base, which I assumed I would do with other sterilizations (fcos)

PR #566, opened 2024-11-26
Stablize fcos v1_6,  base v0_6, and bump all other exp to latest 

prestist:
https://github.com/coreos/butane/issues/559

PR #567, opened 2024-11-27
docs/specs: Fix experimental spec mapping

travier:
The experimental specs are now using the 3.6.0-experimental ignition spec.

Fixes: https://github.com/coreos/butane/pull/566


travier:
I missed that during the review.

prestist (approved):
Ah!, good catch lgtm

PR #569, opened 2024-12-03
Release Butane 0.23.0

yasminvalim:
Write release notes in docs/release-notes.md for releasing Butane 0.23.0. See more: https://github.com/coreos/butane/issues/550

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -4,7 +4,17 @@ nav_order: 9
 
 # Release notes
 
-## Upcoming Butane 0.23.0 (unreleased)
+## Upcoming Butane 0.24.0 (unreleased)
+
+### Features
+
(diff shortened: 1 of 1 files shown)


HuijingHei (approved):
LGTM

ruuda:
There is a tag here on GitHub, but no associated release, while there is for the other tags. Is that intentional?

prestist:
> There is a tag here on GitHub, but no associated release, while there is for the other tags. Is that intentional?

Hey, nice catch! The release is still on the way. The butane release has its packages signed by fedora  and that is a queued request which is external to us. So currently we are waiting on that request, once it's fulfilled the butane release will be finalized upstream. You can see the request here https://github.com/coreos/butane/issues/550#issuecomment-2523538436

PR #596, opened 2025-04-29
Dockerfile: use quay.io as source for Fedora base image

aaradhak:
Switch the builder stage from `registry.fedoraproject.org` to `quay.io/fedora/fedora:42` to align with common usage in container build workflows.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1851


dustymabe (approved):
LGTM

PR #568, opened 2024-11-27
Stabilise openshift 4.18 on Ignition 3.4.0 spec

yasminvalim:
Here’s what this PR aims to achieve:

- [x] Stabilise Openshift 4.18
- [x] Uses Ignition 3.4 specs
- [x] Add specs 4.19.0-experimental
- [x] Remove GRUB password support from OCP 4.18

Fixes: https://github.com/coreos/butane/issues/564

Issue #515, opened 2024-02-16
openshift 4.14 and later stable specs document `grub` section while it is not supported


travier:
The GRUB section is documented while it is not supported by the MCO.

AFAIU this section is not translated to Ignition config so this is less of a problem but still, this is not ideal.

travier:
Pending PR: https://github.com/coreos/butane/pull/542

yasminvalim:
Hi! Butane is being merged into the Ignition repository, and the Butane repo will be archived soon.

We opened a new issue to continue this discussion in Ignition: https://github.com/coreos/ignition/issues/2271

Please use that issue for any follow-ups. Thank you!

PR #605, opened 2025-05-14
Stabilize openshift v4.19.0 on 3.5.0

prestist:
Working with @marmijo  to stabilize OCP 4.19 and add OCP 4.20.

Resolves: https://github.com/coreos/butane/issues/604


travier:
Looks OK as a good start but it's missing the 4.20-exp includes and entries.

prestist:
Yeah the remaining steps are 

- Add 4.20 doc entries
- Generate docs
- Add release note entry

marmijo:
> Yeah the remaining steps are
>- Add 4.20 doc entries
>- Generate docs
>- Add release note entry

I just pushed two commits that should take care of these last items.

internal/doc/butane.yaml
```diff
@@ -328,9 +328,9 @@ root:
               replacement: "Supported values are $1, `s390x-eckd`, `s390x-virt`, `s390x-zfcp`, and `x86_64`."
               if:
                 - variant: fcos
-                  min: 1.6.0
+                  min: 1.6.0-experimental
```
prestist, line 331: hmm,I think this stays as 1.6.0

internal/doc/butane.yaml
```diff
@@ -371,7 +371,7 @@ root:
               replacement: "Unsupported"
               if:
                 - variant: openshift
-                  max: 4.18.0
+                  max: 4.19.0
```
prestist, line 374: stays 4.18

internal/doc/butane.yaml
```diff
@@ -380,7 +380,7 @@ root:
                   replacement: "Unsupported"
                   if:
                     - variant: openshift
-                      max: 4.18.0
+                      max: 4.19.0
```
prestist, line 383: Stays 4.18

internal/doc/butane.yaml
```diff
@@ -390,7 +390,7 @@ root:
...
```
prestist, line 393: Stays 4.18

prestist (commented):
Thank you for taking this on, just a few small changes.

internal/doc/butane.yaml
marmijo: fixed!

internal/doc/butane.yaml
```diff
@@ -328,9 +328,9 @@ root:
               replacement: "Supported values are $1, `s390x-eckd`, `s390x-virt`, `s390x-zfcp`, and `x86_64`."
               if:
                 - variant: fcos
-                  min: 1.6.0
+                  min: 1.6.0-experimental
                 - variant: openshift
-                  min: 4.19.0
+                  min: 4.20.0-experimental
```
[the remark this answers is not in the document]
marmijo: done!

travier:
The regenerate test says that the docs need to be regenerated but the diff for GRUB does not look correct. It should still say unsupported.

docs/upgrading-openshift.md
```diff
@@ -13,6 +13,34 @@ Occasionally, changes are made to OpenShift Butane configs (those that specify `
 1. TOC
 {:toc}
 
+## From Version 4.18.0 to 4.19.0
+
+There are no breaking changes between versions 4.18.0 and 4.19.0 of the `openshift` configuration specification. Any valid 4.18.0 configuration can be updated to a 4.19.0 configuration by changing the version string in the config.
+
+### Luks cex
```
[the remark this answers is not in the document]
travier: I think this should be the same as the FCOS docs unless there are specific changes needed?

docs/upgrading-openshift.md
```diff
@@ -13,6 +13,34 @@ Occasionally, changes are made to OpenShift Butane configs (those that specify `
...
+    - rd.luks.key=/etc/luks/cex.key
+storage:
+  luks:
+    - name: "luks-cex"
+      device: "/dev/sdb"
+      discard: true
+      cex:
+        enabled: true
+```
+
```
travier, line 43: We should also likely add the section about the new boot layouts here: https://github.com/coreos/butane/blob/main/docs/upgrading-fcos.md#boot_device-layouts-s390x-support

travier:
https://github.com/coreos/butane/issues/607

docs/upgrading-openshift.md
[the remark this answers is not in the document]
travier: Ah, I see it in https://coreos.github.io/ignition/migrating-configs/#from-version-340-to-350, but in this case this is the Ignition config and here it's the sugar, so the karg should not be needed?

travier:
@madhu-pillai https://coreos.github.io/ignition/migrating-configs/#from-version-340-to-350 adds a kernel argument but https://coreos.github.io/butane/upgrading-fcos/#luks-cex-support does not and it is not added by the sugar as far as I can see. Am I missing something? Is this karg needed? Which part is correct?

config/openshift/v4_20_exp/translate.go
```diff
@@ -12,14 +12,14 @@
...
 
 import (
 	"net/url"
 	"strings"
 
 	"github.com/coreos/butane/config/common"
-	"github.com/coreos/butane/config/openshift/v4_19_exp/result"
+	"github.com/coreos/butane/config/openshift/v4_20_exp/result"
 	cutil "github.com/coreos/butane/config/util"
 	"github.com/coreos/butane/translate"
```
travier, line 24: s/ToMachineConfig4_19Unvalidated/ToMachineConfig4_20Unvalidated/ later in this file.

A comment needs an update as well.

travier: ToMachineConfig4_19Unvalidated + comment as well need an update

travier: ToConfigBytes 4.17 -> 4.20

travier:
The 4.19 docs from [`beea431` (#605)](https://github.com/coreos/butane/pull/605/commits/beea431fc8ef0e8cc87b89094e50f0f4591fb854) should probably be another previous commit.

travier:
For the commit order, let's do all of the 4.19 changes then the 4.20 ones, or the reverse, not a mix as that makes things harder to follow.

Either like https://github.com/coreos/butane/pull/547/commits or https://github.com/coreos/butane/pull/568/commits.

travier:
Ideally each commit should build and pass CI.

prestist:
> kernel_arguments:
>   should_exist:
>     - rd.luks.key=/etc/luks/cex.key

Hmm so then without the kernel argument cex will not function @madhu-pillai ?
If so then maybe we should add sugar to do this via butane? 

>however in the redhat documentation we did not use the sugar part

wait is there existing sugar?

marmijo:
@travier are there any blockers here, or do you think we're good to review and merge?

travier:
Where is the docs for that in OpenShift? If we change this now this is really late in the release cycle so this is a big risk.

travier:
I don't understand why we need this karg here as the key is hardcoded here: https://github.com/coreos/fedora-coreos-config/pull/2986/files#diff-c526dfd1e153e193423394e0ed643fb22184d4d45167d1737754e8564c3f5becR141

docs/upgrading-openshift.md
```diff
@@ -13,6 +13,92 @@ Occasionally, changes are made to OpenShift Butane configs (those that specify `
 1. TOC
 {:toc}
 
+## From Version 4.18.0 to 4.19.0
+
+There are no breaking changes between versions 4.18.0 and 4.19.0 of the `openshift` configuration specification. Any valid 4.17.0 configuration can be updated to a 4.18.0 configuration by changing the version string in the config.
```
travier, line 18:
```suggestion
There are no breaking changes between versions 4.18.0 and 4.19.0 of the `openshift` configuration specification. Any valid 4.18.0 configuration can be updated to a 4.19.0 configuration by changing the version string in the config.
```

docs/specs.md
```diff
@@ -26,6 +26,8 @@ We recommend that you always use the latest **stable** specification for your op
   - [v1.1.0](config-flatcar-v1_1.md)
   - [v1.0.0](config-flatcar-v1_0.md)
 - OpenShift (`openshift`)
+  - [v4.19.0](config-openshift-v4_19.md)
+  - [v4.18.0](config-openshift-v4_18.md)
```
travier, line 30: As we'll have to update this, can we split the 4.18 fix in another commit?

travier:
Let's add the kargs to the FCOS & OCP examples in the migration docs.

madhu-pillai:
@travier  , As our earlier discussion regards to the CEX feature. You suggestd that key can be passed  as `kargs`, so that on occasion of reboot cryptsetup finds the key file to decrypt the device. Otherwise it waits for the passphrase. eg I just install with ignition without passing the `rd.luks.key=/etc/luks/cex.key` it did the encryption and boots up , but when we reboot it ask for passphrase.
```

[  OK  ] Reached target Preparation for Local File Systems.
[  OK  ] Reached target Local File Systems.
[    3.218573] systemd[1]: Reached target Preparation for Local File Systems.
[    3.218599] systemd[1]: Reached target Local File Systems.
[    3.325622] systemd[1]: Started Forward Password Requests to Clevis.
[    3.325976] systemd[1]: Started Dispatch Password Requests to Console.
🔐 Please enter passphrase for disk root: (press TAB for no echo) [    3.346847] systemd-tty-ask-password-agent[714]: Starting password query on /dev/ttysclp0.
```

docs/specs.md
marmijo: done!

marmijo:
> Let's add the karg for CEX support to the FCOS & OCP examples in the migration docs.

I re-added it to the upgrading-ocp example and included a separate commit for the upgrading-fcos example.

docs/specs.md
```diff
@@ -52,7 +52,7 @@ Do not use **experimental** specifications for anything beyond **development and
 - Flatcar (`flatcar`)
   - [v1.2.0-experimental](config-flatcar-v1_2-exp.md)
 - OpenShift (`openshift`)
-  - [v4.19.0-experimental](config-openshift-v4_19-exp.md)
```
travier, line 54: Nit: this should only add the 4.20 one here.

travier (commented):
Looks like we forgot to remove GRUB support as well: https://github.com/coreos/butane/issues/515

I'll update this PR.

travier:
@madhu-pillai Why is https://github.com/openshift/openshift-docs/pull/93101/files#diff-dbb20b8a9af3ad94e88f5253197aba298510d4bfdc7bff67c207a824b3805b96R48 using a different config example than https://github.com/coreos/butane/blob/main/docs/upgrading-fcos.md#luks-cex-support ? Shouldn't we be using the `boot_device` + `layout` types there as well?

madhu-pillai:
> @madhu-pillai Why is https://github.com/openshift/openshift-docs/pull/93101/files#diff-dbb20b8a9af3ad94e88f5253197aba298510d4bfdc7bff67c207a824b3805b96R48 using a different config example than https://github.com/coreos/butane/blob/main/docs/upgrading-fcos.md#luks-cex-support ? Shouldn't we be using the `boot_device` + `layout` types there as well?

Oh . In the RH documentation , i just followed the NBDE procedure for s390x.  my assumption was both way we can use so i followed the existing NBDE which is bit explicit or the user could use the shortcut from the butane. I did not realize that butane and RH docs to be in sync.

prestist:
Thank you for the edits @travier, taking a quick look over them.

travier (approved):

travier:
LGTM. Let's merge that. Let's hold on the release however until we can get https://github.com/coreos/butane/pull/610 in, or decide not to do it. It will depend on the discussion regarding the docs changes. At least as this, this PR is correct.

marmijo:
Thanks for the last minute edits @travier!

Issue #607, opened 2025-05-16
[ocp 4.14] Likely missing upgrade notes from the update to fcos 1.5 spec


travier:
Most of the update mentions from the FCOS 1.5.0 update are missing in the OCP docs for 4.14: https://github.com/coreos/butane/blob/main/docs/upgrading-fcos.md#from-version-140-to-version-150

We don't need the GRUB password one, but the rest should likely be there?

PR #608, opened 2025-05-16
docs/upgrading-openshift: add examples for features intoduced in 4.14

prestist:
The transition from 4.13 to 4.14 jumped from fcos 1.3 to 1.5. With that there are a number of features which were introduced into the new spec. Add missing feature examples.

fixes:https://github.com/coreos/butane/issues/607


docs/upgrading-openshift.md
```diff
@@ -60,6 +62,130 @@ passwd:
...
+        - name: example-drop-in.conf
+          contents_local: example.conf
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys_local:
+        - id_rsa.pub
+```
+
+### Automatic generation of systemd swap units
```
travier, line 169: I wonder if we support swap in OCP. Maybe we do now but we did not in the past. Let's check that.

travier (commented):
Thanks. Looks good. One question about swap.

docs/upgrading-openshift.md
prestist: This will not work, good catch; no support for filesystem changes after day-0

yuqi-zhang: This falls under the general case of https://issues.redhat.com/browse/MCO-1002 I believe, where right now it's only allowed at install time

prestist: So to be clear, since this is talking about setting swap at install time this correct to document? @yuqi-zhang

yuqi-zhang: Yes, I was unsure as well because the change is under `docs/upgrading-openshift.md`

If this is install time only, let's make sure we clarify that.

PR #610, opened 2025-05-19
config/fcos/v1_7_exp/translate.go: added required kernel arguments to…

madhu-pillai:
Added the required `kernelArguments` to unlock the device in the s390x sugar.
```
[root@a3elp53 butane]# cat bin/s390x/fcos.yaml 
variant: fcos
version: 1.7.0-experimental
boot_device:
  layout: s390x-eckd
  luks:
    device: /dev/dasda
    cex:
      enabled: true
```

```
[root@a3elp53 s390x]# ./butane fcos.yaml -o fcos.ign
[root@a3elp53 s390x]# cat fcos.ign | jq '.'
{
  "ignition": {
    "version": "3.6.0-experimental"
  },
  "kernelArguments": {
    "shouldExist": [
      "rd.luks.key=/etc/luks/cex.key"
    ]
  },
  "storage": {
    "filesystems": [
      {
        "device": "/dev/mapper/root",
        "format": "xfs",
        "label": "root",
        "wipeFilesystem": true
      }
    ],
    "luks": [
      {
        "cex": {
          "enabled": true
        },
        "device": "/dev/dasda2",
        "label": "luks-root",
        "name": "root",
        "wipeVolume": true
      }
    ]
  }
}
```


docs/release-notes.md
```diff
@@ -9,6 +9,7 @@ nav_order: 9
 ### Features
 
 - Validate merged/replaced Ignition configs if they are local/inline _(all base specifications)_
+- Added Kernel Argument rd.luks.keys in s390x layout sugar.
```
prestist, line 12: Wrong tense, always present not past

"Expand existing `CEX` sugar to include a required kernel arg"

config/fcos/v1_7_exp/translate.go
```diff
@@ -262,6 +262,12 @@ func (c Config) processBootDevice(config *types.Config, ts *translate.Translatio
 				Name:       "root",
 				WipeVolume: util.BoolToPtr(true),
 			}}
+			// Add Kernel Arguments for CEX key file
+			rendered.KernelArguments = types.KernelArguments{
+				ShouldExist: []types.KernelArgument{
+					types.KernelArgument("rd.luks.key=/etc/luks/cex.key"),
```
prestist, line 268: To be 100% sure, the path will always be the same? no chance it would vary between systems?

prestist (commented):
Thank you for working on this, and getting these changes in so quickly! 
The changes make sense, I just have a few nits/questions. 

The commit message exceeds 80 character limit, would you mind making it more concise to abide by that and expand on details in a new line under the summary if necessary? 

Additionally can we add a test verifying the sugar's functionality in the translation tests?

config/fcos/v1_7_exp/translate.go
travier: If I understand correctly, it's hardcoded here: https://github.com/coreos/fedora-coreos-config/pull/2986/files#diff-c526dfd1e153e193423394e0ed643fb22184d4d45167d1737754e8564c3f5becR141

travier: https://github.com/coreos/ignition/pull/1820/files#diff-01035e9282707d3e15d0f2f82a9869e11b017833be34e0589dab4c5282b708c2

travier:
https://github.com/coreos/butane/pull/605#issuecomment-2895239967

travier:
I think it's too late to change this now. We should add validation that makes sure that the karg exists and is set to the correct value instead.

travier:
As this would be validation, this does not block the 4.19 spec stabilisation and can happen later.

travier:
https://github.com/coreos/butane/issues/611

travier:
Closing this one as we'll do validation instea: https://github.com/coreos/butane/issues/611

travier:
I'm re-opening this one as we'll have to edit the OCP docs anyway to fix other things, so we might as well get this one in and re-do all the testing.

travier:
Turns out we should not just add this as a karg for the openshift case as that would add an unnecessary reboot, so we need to special case the openshift translation. See: https://github.com/coreos/butane/issues/613

travier:
I tried working on this a bit and the more it goes, the less I like it. So let's go back to what we had initially decided in https://github.com/coreos/butane/pull/610#issuecomment-2895261748 and do validation only: https://github.com/coreos/butane/issues/611

PR #612, opened 2025-05-21
packit: add initial support

HuijingHei:


prestist:
@HuijingHei I added a comment to the jira, the workflow for packit does not fit well with butane or ignition due to the issue of the bin's been signed by fedora. This requires that the upstream release actually finalizes after downstream which is the reverse of propose downstream functionality

HuijingHei:
Thank you @prestist for the info, then should close this.

PR #630, opened 2025-07-30
Remove special case for LUKS with FIPS

travier:
According to investigations, we never actually needed to explicitely set this algorithm.

On modern RHEL (9.6+, OCP 4.20+), we can rely on cryptsetup selecting the right FIPS compliant algorithms when run in FIPS mode in the initrd, so we can remove this logic entirely.

Only done for openshift 4.20 spec for now.

Reverts: [`2ee1017` (#218)](https://github.com/coreos/butane/pull/218/commits/2ee1017a063f22062f87b4380be287134ff02bb8)
See: https://github.com/coreos/butane/pull/625
Fixes: https://github.com/coreos/rhel-coreos-config/issues/10

---

Related: https://issues.redhat.com/browse/STOR-2358


travier:
Backporting to 4.19 & 4.18 should be fine.

travier:
AFAIU, there are not FIPS-140-2/3 implications so we can do this for 4.20 only. We could also backport it as that should not be an issue but it's a risk.

prestist:
Ugh this is weird though, when this lands its entirely reasonable to be in the situation where someone is deploying to 4.20 or .19 .18 and they are using an older config and not be fips compliant due to the ign file options. But if they ran the convert with a newer version of butane they would be.

travier:
To clarify here, all options are FIPS-140-2 & 140-3 compliant so no system is out of compliance right now regarding this PR.
The new defaults (i.e. the one overridden by the hardcoded values removed here) are "just" more robust algorithms.

jlebon (commented):
This looks sane to me.

Re. backporting, I agree with Steven's concerns around changing semantics for existing spec versions. Though I guess we could add a new e.g. v4.19.1 version? I don't think we've used the patch version for specs before so that would be unprecedented (it's also extremely confusing that it's not actually about the v4.19.1 OCP release).

It's of course also possible to make this a docs item instead for 4.19.

travier:
Agree with the backporting concerns. I was thinking of making this a minor spec bump for stable specs as well.

But as Jonathan said, it's confusing and we never did it before so will probably never be used. So let's keep it for 4.20?.

Issue #544, opened 2024-08-26
Set owner for trees.


jeppester:
I'm making a butane config for running nextcloud, based on [this](https://github.com/LorbusChris/nextcloud-quadlet)

If possible I want the containers to run rootless, which I believe is possible if I place the containers in `.config/containers/systemd`.

My plan is to have a `.config`-folder in my repo, that I can simply copy to `/home/coreos` by using the `trees` functionality.

Unfortunately it looks like `trees` does not support setting an owner and will simply default to root.

Do I need to add the files one-by-one to butane in order to set the owner to `coreos`, or is there another way to set the owner for all the files during the install?

PR #644, opened 2025-09-22
Tree permissions

vic1707:
Fixes: https://github.com/coreos/butane/issues/544

Added implementation for specifying custom modes (for files and dirs) as well as the user/group combe for ownership to trees entries


prestist (changes requested):
The documentation was updated for stable variants (v1_6, v1_5, v1_4, etc.) but these variants don't actually have those fields because they use older base schemas.

`*-exp.md` should be the only doc's updated

vic1707:
Thanks for the review on this one, I just squashed everything.

---

You're right about the docs, I blindly ran the generation script after the original CI failed, that's on me.

Restoring the existing files (keeping only the `*-exp.md` ones) currently leads to failing CI (https://github.com/vic1707/butane/actions/runs/19377334435/job/55447945838?pr=5). 

My current best guess is that I modified shared files amongst all versions of butane, so docs generation modifies everything, not just unreleased versions hence CI fail. 
I'll try to look into it and #631 over the weekend

prestist (approved):
LGTM thank you for all your work here @vic1707 !

Issue #592, opened 2025-04-15
aarch64: Mirroring boot disk config should not include `bios` part


HuijingHei:
The setup refer to https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_reconfiguring_the_root_filesystem
```
[coreos-assembler]$ rpm -q butane
butane-0.23.0-1.fc40.aarch64

[coreos-assembler]$ arch 
aarch64

[coreos-assembler]$ cat test-raid1.bu 
variant: fcos
version: 1.6.0
boot_device:
  mirror:
    devices:
      - /dev/vda
      - /dev/vdb

[coreos-assembler]$ butane -p test-raid1.bu -o test-raid1.ign                         
[coreos-assembler]$ cat test-raid1.ign                                                
{                             
  "ignition": {               
    "version": "3.2.0"        
  },                                                                                  
  "storage": {                                                                                                                                             10:24:54 [83/1959]
    "disks": [
      {
        "device": "/dev/vda",
        "partitions": [
          {
            "label": "bios-1",
            "sizeMiB": 1,
            "typeGuid": "21686148-6449-6E6F-744E-656564454649"
          },
...
        ],
        "wipeTable": true
      },
      {
        "device": "/dev/vdb",
        "partitions": [
          {
            "label": "bios-2",
            "sizeMiB": 1,
            "typeGuid": "21686148-6449-6E6F-744E-656564454649"                                                                                             10:24:54 [52/1959]
          },
...
        ],
        "wipeTable": true
      }                                                                                                                                                    10:24:54 [36/1959]
    ],
...
    ]
  }
}
```

travier:
Hum, could you clarify what the issue is? The resulting Ignition config should not include the `bios-1/2` partitions?

HuijingHei:
> Hum, could you clarify what the issue is? The resulting Ignition config should not include the `bios-1/2` partitions?

Yes, as aarch64 is using EFI so do not need `bios-` part, or with the mirror setup, during boot will get error:

```
Displaying logs from failed units: ignition-ostree-transposefs-save.service
-- Logs begin at Tue 2025-04-15 08:43:50 UTC, end at Tue 2025-04-15 08:44:13 UTC. --
Apr 15 08:44:11 ignition-ostree-transposefs[1066]: Moving bootfs to RAM...
[...]

PR #653, opened 2025-11-04
bugfix: add validation for bootdevice mirror

prestist:
It comes down to we should have made it a requirement when using device.mirror.
To not break existing configs lets add a warning from start, that way people will be notified but not interrupted. 

fixes: https://github.com/coreos/butane/issues/592

### DIFF
--- a/config/common/errors.go
+++ b/config/common/errors.go
@@ -57,6 +57,7 @@ var (
 	ErrUnknownBootDeviceLayout       = errors.New("layout must be one of: aarch64, ppc64le, s390x-eckd, s390x-virt, s390x-zfcp, x86_64")
 	ErrUnknownBootDeviceLayoutLegacy = errors.New("layout must be one of: aarch64, ppc64le, x86_64")
 	ErrTooFewMirrorDevices           = errors.New("mirroring requires at least two devices")
+	ErrMirrorRequiresLayout          = errors.New("boot_device.layout should be specified when boot_device.mirror is specified")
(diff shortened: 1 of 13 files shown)


prestist:
@travier Hmm, from my understanding the x86 layout would always work. but if in this case aarch and the layout is not specified then it would not work and result in what @HuijingHei ran into. 

Initially I was going to make it an error, but wanted to effect stable versions because this would be true for all of boot_device's existence 1.3+ so I made it a warning. So tbc it works for some layouts i.e x86 without specification but others it will just fail like aarch.

Do we want to change this over to an error?

prestist:
Yeah absolutely easy! 
Thank you!

prestist:
Ok I was working on this and just clicked. 
How would we know that we are on a version that is not x86 unless its specified via a layout? i.e we would not ever get to an error state.

`make it an error for all versions for non x86, as this is already failing so we are not breaking things that work.`

I think we can do this =>

make it a warning for all stable specs
make it an error (and update our examples in the docs) for the latest 1.7 exp spec and mark it as an important change in the release notes

travier:
Ah indeed, we don't know the arch if it's not there. Then I agree with your latest proposal.

prestist:
I think these changes should be gtg.

I did not add any examples with 'or the latest 1.7 exp' as I dont think that has been convention to add examples of exp specs. 

It might be a different convo to have a separate examples.md for up and coming features but I dont think we should mix stable examples with experimental.

yasminvalim (approved):

PR #656, opened 2025-11-16
chore: package dependencies debt

vic1707:
PR implementing: https://github.com/coreos/butane/pull/631#issuecomment-3529809716 in order to fix the circular dependency of #631.

From @prestist
> ### New Structure:
> ```
> translate/              ← NEW: Top-level package
>   interface.go          ← Define Translator interface
>   registry.go           ← Registry implementation
> 
> base/v0_7_exp/
>   schema.go             ← Data structures only
>   translator.go         ← Implements translate.Translator interface
> 
> config/fcos/v1_7_exp/
>   schema.go             ← Wraps base schemas
>   translator.go         ← Implements translate.Translator interface
>   init.go               ← Registers with translate.Registry
> ```
> [...]
> So I would recommend that we do it in stages,
> 
> 1. define the interfaces.
> 2. review that
> 3. Implement the use of the interfaces

### DIFF
--- a/translator/interface.go
+++ b/translator/interface.go
@@ -15,8 +15,6 @@
 package translator
 
 import (
-	"context"
-
 	"github.com/coreos/vcontext/report"
 )
 
@@ -27,11 +25,10 @@ import (
 type Translator interface {
 	// Metadata the variant, version, and target Ignition version.
 	Metadata() Metadata
-
-	// Translate converts Butane config bytes to Ignition config bytes.
-	Translate(ctx context.Context, input []byte, opts Options) (Result, error)
-
-	// Validate validates a Butane config without performing translation.
-	Validate(ctx context.Context, input []byte) (report.Report, error)
+	// Parse yml into schema struct, basically a yaml.Unmarshal wrapper?
+	Parse(input []byte /*opts?*/) (interface{}, error)
+	// From inner schema struct to Ignition struct
+	Translate(input interface{}, options Options) (interface{}, report.Report, error)
+	// Validates yml inner struct
+	Validate(in interface{}) (report.Report, error)
 }
-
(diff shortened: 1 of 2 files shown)


prestist:
@vic1707 No worries on the pings! Looking now.

prestist (commented):
Ah, I am sorry, yes there is an interface but we need it to do more for us. Currently its being used for "translate" only annd because of how its used we cannot use it in a way that enables us to lookup the implementation we need at runtime.

prestist:
@vic1707 No worries, and there is no rush on this. I will take a look today and see what I can do to help :)

prestist:
@vic1707 I gave it a shot, there is likely something I am overlooking but this is my interpretation of what we need. 

You can see the registry is basically a map, which we can register all of our implementations on.

The next step would be to implement this interface and see where the holes are, I would probably try it with base07_exp config and lets see.

vic1707:
happy I pushed back a little 😄 
I should get some free time over the weekend to implement one translator, we'll see how this goes

vic1707:
Hi, sorry to everyone waiting on me here, finally found some time to work here again.
Just pushed a WIP impl on fcos 1.8 with the new interface.

@prestist may I ask you to review that? Knowing the other implementations will be very similar for all other backends getting this one right should be enough to move forward (minus OpenShift which will depend on Options.Raw)


Note: TranslationSet was removed from public Result, now considered an implementation detail

yasminvalim:
@vic1707 thanks for your work here!

Please, feel free to open a PR and continue to work on this issue in Ignition. I will be glad to review. 🐱

vic1707:
continued in https://github.com/coreos/ignition/pull/2300

PR #688, opened 2026-02-27
Add cli option to output all strings in plain text

### LINKED ISSUE #670: Add cli option to output all strings in plain text in the Ignition config
Always output inlined or other `source` in plaintext if an option is passed to allow inspecting the `.ign` file more easily.

Reference #94 to see where the compression was added, and what files can be modified to undo it on feature flag.

Phuong-Le-NN:
Adding --plain-text CLI flag to generate an Ignition config in plain text This generated config should be used for debugging only

Fixes: #670

### DIFF
--- a/base/util/url.go
+++ b/base/util/url.go
@@ -24,7 +24,13 @@ import (
 	"github.com/vincent-petithory/dataurl"
 )
 
-func MakeDataURL(contents []byte, currentCompression *string, allowCompression bool) (uri string, compression *string, err error) {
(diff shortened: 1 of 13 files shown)


travier:
Can you turn `allowCompression` into an enum and base the logic on that instead? That would having 2 booleans controlling some parts of the same logic in the code.

I think we can also enable that functionality for all versions as this is only for debugging and should apply equally.

yasminvalim:
@Phuong-Le-NN  thanks for your work here!

Please, feel free to open a PR and continue to work on this issue in Ignition if makes sense. I will be glad to review. 🐱

PR #692, opened 2026-03-18
opencode/skills/*

prestist:
Lets start adding opencode skills for common tasks, to enable ease of use on common tasks.


yasminvalim (approved):
/LGTM

PR #694, opened 2026-03-20
Quadlet sugar

angelcerveraroldan:
Basic sugar for quadlets. Allows for the following syntax:

```
variant: fcos
version: 1.8.0-experimental
systemd:
  quadlets:
    - name: sleepy-disk.container
      rootful: true
      contents_local: containers/sleepy.container
    - name: sleepy-infinity.container
      rootful: true
      contents: |
        [Unit]
        Description=A sleepy container
        [Container]
        ContainerName=sleepy-pod-inf
        Image=quay.io/fedora/fedora
        Exec=sleep infinity
        [Install]
        WantedBy=multi-user.target
```

I still need to add examples for the docs, and to run more tests.

Docs about quadlets can be found [here](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html).

### DIFF
--- a/base/v0_8_exp/translate.go
+++ b/base/v0_8_exp/translate.go
@@ -100,9 +100,13 @@ func (c Config) ToIgn3_7Unvalidated(options common.TranslateOptions) (types.Conf
 
 	c.addMountUnits(&ret, &tm)
 
-	tm2, r2 := c.processTrees(&ret, options)
-	tm.Merge(tm2)
-	r.Merge(r2)
+	tmTrees, rTrees := c.processTrees(&ret, options)
+	tmQuadlets, rQuadlets := c.processQuadlets(&ret, options)
+
+	tm.Merge(tmTrees)
+	tm.Merge(tmQuadlets)
+	r.Merge(rTrees)
+	r.Merge(rQuadlets)
 
 	if r.IsFatal() {
 		return types.Config{}, translate.TranslationSet{}, r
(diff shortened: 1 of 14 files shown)


base/v0_8_exp/translate.go
```diff
@@ -297,6 +301,172 @@ func translateDropin(from Dropin, options common.TranslateOptions) (to types.Dro
...
+	t := newNodeTracker(ret)
+	quadletsPath := path.New("yaml", "systemd", "quadlets")
+	ts.AddTranslation(quadletsPath, path.New("json", "storage")) // quadlets will be translated to storage (files and links)
+	for quadletNum, quadlet := range c.Systemd.Quadlets {
+		quadletPath := quadletsPath.Append(quadletNum)
+
+		// We need to handle `foo@bar.container` differently than `foo@.container`, as the former needs to be a symlink to the latter
+		var tsFile translate.TranslationSet
+		var rFile report.Report
+		if isTemplateInstance, templateName := isTemplateInstance(quadlet.Name); isTemplateInstance {
```
prestist: nit: the variable isTemplateInstance shadows the function name, its not a bug but its kinda confusing to read, maybe rename it to "isInstance" or something similar.

base/v0_8_exp/translate.go
```diff
@@ -297,6 +301,172 @@ func translateDropin(from Dropin, options common.TranslateOptions) (to types.Dro
 	return
 }
 
+// buildQuadletPath returns the filesystem path for a quadlet.
+// See https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html
+func buildQuadletPath(isRoot bool, quadletName string) string {
+	const (
+		adminContainersPath = "/etc/containers/systemd"
+		userContainersPath  = "/etc/containers/systemd/users"
```
prestist, line 309: Just want to make sure this is intentional -- per the [podman docs](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html), `/etc/containers/systemd/users/` makes the quadlet run for **all** non-root users. If the intent is to target a specific user it would need to be `/etc/containers/systemd/users/${UID}`.

For ignition this is probably the right default, but might be worth calling out in the docs that `rootful: false` means "all users" and not a specific user.

base/v0_8_exp/translate.go
```diff
@@ -297,6 +301,172 @@ func translateDropin(from Dropin, options common.TranslateOptions) (to types.Dro
...
+}
+
+// isTemplateInstance checks if a quadlet name is a template instance (e.g. foo@100.container).
+// Returns true and the base template name (e.g. foo@.container) if it is an instance.
+func isTemplateInstance(name string) (bool, string) {
+	splitIndex := strings.Index(name, "@")
+	if splitIndex == -1 {
+		return false, ""
+	}
+	extensionIndex := strings.LastIndex(name, ".")
```
prestist, line 327: What happens if `name` has no `.` at all (e.g. `foo@bar`)? `LastIndex` returns -1 and then `splitIndex+1 == extensionIndex` would be false, so we'd return `true` with a malformed template name. I know `quadletExtension` validation would catch it before we get here but this feels a bit fragile on its own.

base/v0_8_exp/validate.go
```diff
@@ -91,16 +91,54 @@ func (t Tree) Validate(c path.ContextPath) (r report.Report) {
...
-	return
+
+	return nil
+}
+
+// Validate checks the quadlet name has a valid extension and template instances don't have contents.
+func (rs Quadlet) Validate(c path.ContextPath) (r report.Report) {
+	r = validateNotTooManySources(rs.ContentsLocal, rs.Contents, c)
+	// Template instances cannot have a content as they are symlinks
+	if isTemplateInstance, _ := isTemplateInstance(rs.Name); isTemplateInstance {
```
prestist: Hmm, if a template instance has both `contents` AND `contents_local` set we would get `ErrTooManySystemdSources` from `validateNotTooManySources` and then also `ErrTemplateInstanceCannotHaveContents` here. Both errors are technically correct but might be worth adding a test case for that specific combo to make sure the report looks right.

prestist (commented):
Nice work overall, the approach follows existing patterns well (similar to trees and mount units). The `validateNotTooManySources` refactor is a clean improvement too.

A few things I noticed -- mostly around the user path semantics and some edge cases in the template instance logic. Also would be good to add a test for quadlet collisions with `storage.files` entries to make sure the nodeTracker catches those cross-section conflicts.

I see the PR description mentions still needing examples and more tests, so some of this might already be on your radar. Thank you for working on this!

base/v0_8_exp/translate.go
angelcerveraroldan: Yes, this was by design. I actually was not sure wether we shuld just have `/etc/containers/systemd/` and remove the users one. This would align more closely with ignition which does not save to `/etc/systemd/user/`. I do think that it may be good to keep it, as it is not a lot more code when compared to just having the one path.

Either way, I will have to document this.

base/v0_8_exp/translate.go
angelcerveraroldan: The validation shuld catch this, as it checks for extension, to make the function less fragile, I have added a check for `LastIndex == -1`.

base/v0_8_exp/validate.go
angelcerveraroldan: Ah, didn't think of this. I think if a template has both `contents` and `contents_local`, then `ErrTooManySystemdSources` would give an unhelpful message:

```
only one of the following can be set: contents, contents_local
```

I have changed it so that it just give two errors `ErrTemplateInstanceCannotHaveContents`, one for each of them.

prestist:
Ah thank you for your quick edits! I was about to approve this but one small nit popped up. Can you update your commits to follow the convention of this repo? typically we name a file for folder followed by : and the change. 

Just a small thing overall but its good to keep it consistent  ex: https://github.com/coreos/butane/commit/2bb9e42c56a00980217ea8355ecf3b29b0c1b454!

prestist (commented):
@angelcerveraroldan 

So close with the commit names! 

I think that it would be clear to have both of these commits to be squashed into one commit

Look at the sugar that has been added here https://github.com/coreos/butane/pull/484/changes/0d2fa0fb99bf81a95c6f3c73379bfd8af36d9cba

prestist (approved):
LGTM thank you :)

PR #696, opened 2026-04-14
Pretty error reporting

angelcerveraroldan:
The goal of this PR is to add pretty error reporting, since the difference is visual, I have attached screenshots for the errors that butane would report for the following butane:

```yaml
variant: fcos
version: 1.6.0
boot_device:
  layout: s390x-virt
  mirror:
    devices:
      - /dev/vda
      - /dev/vdb
systemd:
  units:
    - name: example.service
      enabled: true
      contents: |
        [Unit]
        Description=Example
      contents_local: example.service
    - name: other.service
      dropinz:
        - name: override.conf
          contents: |
            [Service]
            ExecStart=
    - name: valid.service
      enabled: true
```

How errors show without this feature:

current-errors

How errors show with the new feature:

new-errors

# New flags added:

1. `--pretty-errors` to turn on pretty error reporting
2. `--color[=WHEN]`, where WHEN is auto (default), always, or never. This follows the same convention as `ls`, `ripgrep`, `diff`, `bat`, `exa`,  ... In auto mode, colors are disabled if stderr is not a terminal or if the [NO_COLOR](https://no-color.org/) environment variable is set. It also accepts `colour` just in case

This is an early PR to get feedback on whether or not this is a good addition to butane, or if it should go somewhere else / nowhere. I am not sure if the current flag name `pretty-errors` is good or not, seems kind of verbose to me.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/coreos/butane/config"
 	"github.com/coreos/butane/config/common"
+	breport "github.com/coreos/butane/internal/report"
 	"github.com/coreos/butane/internal/version"
 )
 
@@ -31,14 +32,25 @@ func fail(format string, args ...interface{}) {
 	os.Exit(1)
 }
 
+func isCharDevice(f *os.File) bool {
+	stat, err := f.Stat()
+	if err != nil {
+		return false
+	}
+	return stat.Mode()&os.ModeCharDevice != 0
+}
+
 func main() {
 	var (
 		input       string
 		output      string
+		colorFlag   string
 		check       bool
 		strict      bool
 		helpFlag    bool
 		versionFlag bool
+		rawErrors   bool
+		colorize    bool
 	)
 	options := common.TranslateBytesOptions{}
 	pflag.BoolVarP(&helpFlag, "help", "h", false, "show usage and exit")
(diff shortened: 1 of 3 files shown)


internal/main.go
```diff
@@ -49,6 +61,12 @@ func main() {
 	pflag.BoolVarP(&strict, "strict", "s", false, "fail on any warning")
 	pflag.BoolVarP(&options.Pretty, "pretty", "p", false, "output formatted json")
 	pflag.BoolVarP(&options.Raw, "raw", "r", false, "never wrap in a MachineConfig; force Ignition output")
+	pflag.BoolVar(&prettyErrors, "pretty-errors", false, "pretty print errors")
```
prestist: hmm, I think opt in makes sesne, but having the many different flags can lead to a little confusion. separate flags imply they do something separately, i.e --color without --pretty-errors does not really do anything right?

Is there a way we can make --color a sub-option of --pretty-errors?

travier (approved):
This looks great to me. +1 from me. Untested.

travier:
> --pretty-errors to turn on pretty error reporting

Should we have this on by default? Butane output should be for humans so good default for humans sounds like a good idea.

angelcerveraroldan:
>  --pretty-errors to turn on pretty error reporting
> 
> Should we have this on by default? Butane output should be for humans so good default for humans sounds like a > good idea.

I did not enable it by default in case the raw output was being used in any automatic things. But if its meant for humans only, then having it by default may be nice, as otherwise a lot of people may not know it is available. Open to going either way.

internal/main.go
angelcerveraroldan: The reason I had setup color with those flags is to follow the standard convention that most programs use.

I do think that having colors only be used for pretty errors can be strange, specially so if pretty errors is opt-in rather than opt-out. I am not entirely sure how to make colors a sub option of pretty errors nicely.

Maybe something like: `--pretty-errors=color`, `--pretty-errors=no-color`, as well as `--pretty-errors=auto`, where auto checks if its a tty or not / if NO_COLOR is set. (If --pretty-errors is not set then use the normal errors)

I am not sure if this is less confusing or not.

travier: I would default to the new pretty output and add another option (`--json` ?) to output machine readable output instead.

travier: or `--raw-errors` to use the current format and use that in tests for now (as moving to JSON would likely require us to update tests).

prestist: Ah, I like that @travier, if we just flip it its much cleaner :) @angelcerveraroldan

angelcerveraroldan: Great, I have made the change, now it defaults to pretty errors, and you can change that by using `--raw-errors`

prestist (approved):
Thank you for all your work on this @angelcerveraroldan, it lgtm :)

PR #699, opened 2026-04-16
Fix markdown lint warnings

Nemric:
This PR to fix some markdown lint warnings that won't change docs rendering

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -16,7 +16,6 @@ nav_order: 9
 
 ### Docs changes
 
-
 ## Butane 0.27.0 (2026-02-27)
 
 ### Features
@@ -31,7 +30,6 @@ nav_order: 9
 ### Docs changes
 
 - Re-vendor latest ignition release; 3.6.0-experimental becomes 3.6.0
-
 
 ## Butane 0.26.0 (2026-01-16)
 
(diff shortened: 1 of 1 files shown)


prestist (commented):
This lgtm, I would prefer if we update the commit message to conform with the history 

Maybe something like "docs/release-notes: fix markdown lint warnings" 

@Nemric Ah thank you for that ; I think it would make sense to add CI to prevent drift in the future, I will take a look at adding that in [repo-templates](https://github.com/coreos/repo-templates/)

Nemric:
> This lgtm, I would prefer if we update the commit message to conform with the history
> 
> Maybe something like "docs/release-notes: fix markdown lint warnings"
> 
> @Nemric Ah thank you for that ; I think it would make sense to add CI to prevent drift in the future, I will take a look at adding that in [repo-templates](https://github.com/coreos/repo-templates/)

If you want to add a CI linter, take care to set `"MD024": {"siblings_only": true}` 
see : https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md024---multiple-headings-with-the-same-content

prestist (approved):

prestist:
Thank you for the quick edit!

PR #701, opened 2026-05-08
Dockerfile: update to Fedora 44

marmijo:
xref: https://github.com/coreos/fedora-coreos-tracker/issues/2055


Dockerfile
prestist:
```suggestion
FROM quay.io/fedora/fedora-minimal:44
```

prestist (commented):
We do need to update that to 44 too right?

Dockerfile
marmijo: Ah, I missed this. It's been updated, thanks!

travier (approved):

PR #704, opened 2026-05-19
OCP 4.22 stabilization

prestist:
- Stabilize OpenShift spec 4.22.0, targeting Ignition spec 3.6.0
- Add OpenShift spec 4.23.0-experimental, targeting Ignition spec 3.7.0-experimental
- Remove GRUB config support from OpenShift 4.22 (matching 4.21 pattern)
- Update all documentation (specs, upgrading, release notes)

Closes #702

PR #706, opened 2026-05-19
docs/release-notes: fix openshift quadlets version reference

prestist:
Fix quadlets version reference from `openshift 4.22.0+` to `openshift 4.23.0-exp`
Quadlets only exist in `base/v0_8_exp`, not `base/v0_7` (which `openshift/v4_22` depends on)

Follow-up to #704 per review feedback from @travier


marmijo (approved):
LGTM

PR #707, opened 2026-05-20
Sync repo templates ⚙

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/repo-templates/actions/workflows/sync.yml) ([source](https://github.com/coreos/repo-templates/blob/main/.github/workflows/sync.yml)).

Sync with coreos/repo-templates@5a051c2e5c89e2d4562e321df666bd6396a0c5ef.


travier (approved):

prestist:
@travier LGTM, but I think this is the stage where we should have release notes added stating the updated signing key for in the future wdyt?

travier:
Ah damned  you're right

PR #708, opened 2026-05-20
docs/release-notes: note Fedora 44 signing key

prestist:
Add release notes entry for PR #707, noting that starting with 0.28.0, Butane binaries are signed with the Fedora 44 key


marmijo (approved):
LGTM

PR #711, opened 2026-06-04
fcos: skip ErrReuseByLabel warning for boot_device.mirror disks

### LINKED ISSUE #710: Config.Validate() produces false positive ErrReuseByLabel warnings for boot_device.mirror disks
*Disclaimer: Issue generated by AI with guidance from @dustymabe*

### Description

When a Butane config uses `boot_device.mirror` and also specifies `storage.disks` entries for the same devices (e.g., to override partition sizes or add extra partitions), `Config.Validate()` warns:

```
warning at $.storage.disks.0.partitions.0.number: partitions cannot be reused by label;
number must be specified except on boot disk (/dev/disk/by-id/coreos-boot-disk) or when
wipe_table is true
```

These warnings are false positives. `processBootDevice()` in `translate.go` auto-generates `wipe_table: true` for mirror disks, but `Config.Validate()` runs *before* translation, so it doesn't see the auto-generated `wipe_table`. The user can't reasonably suppress this – adding explicit `wipe_table: true` is redundant, and adding explicit partition `number` fields would break the label-based merge with the auto-generated mirror partitions.

### Reproducer

```yaml
variant: fcos
version: 1.7.0
boot_device:
  layout: x86_64
  luks:
    tpm2: true
  mirror:
    devices:
      - /dev/sda
      - /dev/sdb
storage:
  disks:
    - device: /dev/sda
      partitions:
        - label: root-1
          size_mib: 10240
        - label: var-1
    - device: /dev/sdb
      partitions:
        - label: root-2
          size_mib: 10240
        - label: var-2
  raid:
    - name: md-var
      level: raid1
      devices:
        - /dev/disk/by-partlabel/var-1
        - /dev/disk/by-partlabel/var-2
  luks:
    - name: var
      device: /dev/md/md-var
  filesystems:
    - device: /dev/mapper/var
      path: /var
      label: var
      format: xfs
      wipe_filesystem: true
      with_mount_unit: true
```

This is the "Encrypted mirrored boot disk with separate /var" example from the [FCOS storage docs](https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_advanced_examples) (with `layout: x86_64` added for the separate v1.7 requirement).

### Proposed fix

In `Config.Validate()` (affects v1_6, v1_7, v1_8_exp), skip the `ErrReuseByLabel` check for disks that are listed in `boot_device.mirror.devices`, since `processBootDevice()` will set `wipe_table: true` for them:

```go
func (conf Config) Validate(c path.ContextPath) (r report.Report) {
	mirrorDevices := make(map[string]bool)
	for _, dev := range conf.BootDevice.Mirror.Devices {
		mirrorDevices[dev] = true
	}

	for i, disk := range conf.Storage.Disks {
		if disk.Device != rootDevice && !util.IsTrue(disk.WipeTable) && !mirrorDevices[disk.Device] {
			for p, partition := range disk.Partitions {
				if partition.Number == 0 && partition.Label != nil {
					r.AddOnWarn(c.Append("storage", "disks", i, "partitions", p, "number"), common.ErrReuseByLabel)
				}
			}
		}
	}
	// ...
```

### Context

Reported via https://github.com/coreos/fedora-coreos-docs/issues/808

dustymabe:
Config.Validate() warns when a partition on a non-boot disk uses a label without an explicit number and wipe_table is not set.  However, when the disk is listed in boot_device.mirror.devices, this warning is a false positive: processBootDevice() in translate.go will auto- generate wipe_table: true for mirror disks, but Validate() runs before translation so it doesn't see the auto-generated wipe_table.

The user can't reasonably suppress this: adding explicit wipe_table: true is redundant, and adding explicit partition number fields would break the label-based merge with the auto-generated mirror partitions.

Skip the ErrReuseByLabel check for disks that are listed in boot_device.mirror.devices.

Fixes: https://github.com/coreos/butane/issues/710

Assisted-by: <anthropic/claude-opus-4.6>

### DIFF
--- a/config/fcos/v1_6/validate.go
+++ b/config/fcos/v1_6/validate.go
@@ -35,8 +35,14 @@ var sdRe = regexp.MustCompile("(/dev/sd[a-z]$)")
 // We can't define a Validate function directly on Disk because that's defined in base,
 // so we use a Validate function on the top-level Config instead.
 func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	// Collect mirror device paths so we can skip the reuse-by-label
+	// check for them; processBootDevice() will set wipe_table: true.
+	mirrorDevices := make(map[string]bool)
+	for _, dev := range conf.BootDevice.Mirror.Devices {
+		mirrorDevices[dev] = true
+	}
 	for i, disk := range conf.Storage.Disks {
-		if disk.Device != rootDevice && !util.IsTrue(disk.WipeTable) {
+		if disk.Device != rootDevice && !util.IsTrue(disk.WipeTable) && !mirrorDevices[disk.Device] {
 			for p, partition := range disk.Partitions {
 				if partition.Number == 0 && partition.Label != nil {
 					r.AddOnWarn(c.Append("storage", "disks", i, "partitions", p, "number"), common.ErrReuseByLabel)
--- a/config/fcos/v1_6/validate_test.go
+++ b/config/fcos/v1_6/validate_test.go
@@ -583,6 +583,30 @@ func TestValidateConfig(t *testing.T) {
 				},
 			},
 		},
+		// valid config (disk is a boot_device.mirror device)
+		{
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device: "/dev/sda",
+								Partitions: []base.Partition{
+									{
+										Label: util.StrToPtr("root-1"),
+									},
+								},
+							},
+						},
+					},
+				},
+				BootDevice: BootDevice{
+					Mirror: BootDeviceMirror{
+						Devices: []string{"/dev/sda", "/dev/sdb"},
+					},
+				},
+			},
+		},
 		// invalid config (wipe_table is nil)
 		{
 			in: Config{
--- a/config/fcos/v1_7/validate.go
+++ b/config/fcos/v1_7/validate.go
@@ -36,8 +36,14 @@ var sdRe = regexp.MustCompile("(/dev/sd[a-z]$)")
 // We can't define a Validate function directly on Disk because that's defined in base,
 // so we use a Validate function on the top-level Config instead.
 func (conf Config) Validate(c path.ContextPath) (r report.Report) {
+	// Collect mirror device paths so we can skip the reuse-by-label
+	// check for them; processBootDevice() will set wipe_table: true.
+	mirrorDevices := make(map[string]bool)
+	for _, dev := range conf.BootDevice.Mirror.Devices {
+		mirrorDevices[dev] = true
+	}
 	for i, disk := range conf.Storage.Disks {
-		if disk.Device != rootDevice && !util.IsTrue(disk.WipeTable) {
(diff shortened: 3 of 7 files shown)


travier (approved):
LGTM. Not tested yet.

travier:
This fixes one of the two cases. The second one is:

```
Invalid Butane config at ./modules/ROOT/pages/storage.adoc:92:
  warning at $.storage.disks.0.partitions.0.number, line 8 col 7: partitions cannot be reused by label; number must be specified except on boot disk (/dev/disk/by-id/coreos-boot-disk) or when wipe_table is true
  Config produced warnings and --strict was specified
```

dustymabe:
> ./modules/ROOT/pages/storage.adoc

@travier I think that one is a different problem. In that case we should probably set the partition number.

This is what AI says (see below). It actually recommends setting `wipe_table: true` but I don't think we want to do that here because that would be destructive if people were to reprovision and actually wanted to keep the logs for whatever reason.

---

The second warning is a different case from the `boot_device.mirror` false
positive. It comes from the "Adding /var/log from a secondary disk" example
at line 92 of `storage.adoc`:

```yaml
storage:
  disks:
  - device: /dev/vdb
    wipe_table: false
    partitions:
    - size_mib: 0
      start_mib: 0
      label: log
```

Here `/dev/vdb` is a regular secondary disk -- no `boot_device.mirror` is
involved. The partition has `label: log` but no explicit `number`, and
`wipe_table` is `false`, so `ErrReuseByLabel` fires.

Unlike the mirror case, I think this is a **legitimate warning** and it's the
**docs** that should be fixed. The user's intent is clearly to treat
`/dev/vdb` as a fresh disk (they specify `start_mib: 0` to place the
partition at the beginning), but `wipe_table: false` contradicts that --
it tells Ignition to preserve the existing partition table and try to match
existing partitions.

Two possible docs fixes:

1. **Change `wipe_table: false` to `wipe_table: true`** -- this is the more
   correct fix since the example is treating the disk as fresh anyway.
   Setting `wipe_table: false` with `start_mib: 0` is contradictory intent.

2. **Add `number: 1`** to the partition -- this suppresses the warning but
   doesn't fix the underlying contradiction.

That said, there _is_ a possible butane improvement here too: when a
partition specifies `start_mib` (i.e., the user knows the exact placement),
the warning about "reuse by label" is less relevant since the user isn't
relying on label-matching to find an existing partition -- they're defining
the layout explicitly. But that's a more nuanced validation change and
arguably not worth the complexity.

I'd recommend option 1 (change `wipe_table` to `true`) as a second commit
in the fedora-coreos-docs PR.

travier:
Sounds good indeed. I fixed that in the docs PR.

travier:
Gemini's suggestions looks OK to me. It should not matter but it does not make the code harder to read or understand.

dustymabe:
> Gemini's suggestions looks OK to me. It should not matter but it does not make the code harder to read or understand.

To me it makes it a little harder to read/understand. Since performance/memory use isn't really a concern here I'd prefer to leave it unless you think it is much better.

travier:
Works for me.

PR #717, opened 2026-06-23
OCP 4.22 MCO does not support Ignition spec 3.6.0 produced by Butane openshift 4.22.0

yasminvalim:
As a result when butane stabilized for OCP 4.22  we got into a state where OCP 4.22 renders into ignition 3.6 which the MCO will not know. To work around this for 4.22 we need to do the following.

- [x] **Update upstream (Butane) to have OCP 4.22 point to 3.5 Stable.**
- [x] Create a patch file for, RHAOS, RHEL 8 build, and request a mirror.
- [x] Create a patch for fedora, and build
- [x] Create a patch for C9s and C10s and build.

### DIFF
--- a/docs/config-openshift-v4_21.md
+++ b/docs/config-openshift-v4_21.md
@@ -137,7 +137,7 @@ The OpenShift configuration is a YAML document conforming to the following speci
         * **_needs_network_** (boolean): whether or not the device requires networking.
     * **_cex_** (object): describes the IBM Crypto Express (CEX) card configuration for the luks device.
       * **_enabled_** (boolean): whether or not to enable cex compatibility for luks. If omitted, defaults to false.
-  * **_trees_** (list of objects): a list of local directory trees to be embedded in the config. Symlinks must not be present. Ownership, file modes (using `file_mode`) and directories modes (using `dir_mode`) can be specified for the tree. If not specified, ownership is not preserved and file modes are set to 0755 if the local file is executable or 0644 otherwise. File attributes can be overridden by creating a corresponding entry in the `files` section; such entries must omit `contents`.
+  * **_trees_** (list of objects): a list of local directory trees to be embedded in the config. Symlinks must not be present. Ownership is not preserved. File modes are set to 0755 if the local file is executable or 0644 otherwise. File attributes can be overridden by creating a corresponding entry in the `files` section; such entries must omit `contents`.
     * **local** (string): the base of the local directory tree, relative to the directory specified by the `--files-dir` command-line argument.
     * **_path_** (string): the path of the tree within the target system. Defaults to `/`.
 * **_systemd_** (object): describes the desired state of the systemd units.
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,7 @@ nav_order: 9
 
 ### Misc. changes
 
+- Roll back to Ignition spec 3.5.0 _(openshift 4.22.0)_
 - Add support for pretty error reporting, can be controlled through
   the use of `--raw-errors` (disable) and `--color`/`--colour`
 
--- a/internal/doc/butane.yaml
+++ b/internal/doc/butane.yaml
@@ -39,7 +39,7 @@ mode:
       replacement: are not supported
       if:
         - variant: openshift
-          min: 4.22.0
+          min: 4.23.0
 
 append-contents-local:
   # Mention contents_local on specs that support it.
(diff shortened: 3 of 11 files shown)


prestist (commented):
Thanks for working on this @yasminvalim! The rollback approach follows the established pattern (we've done this for 4.19, 4.17, 4.14, etc.) and the code changes are mechanically correct.

 A couple things to address before merging:

1. The commit messages need to follow the repo's <component>: <description> convention. I'd suggest squashing into a single commit with something like: 
>  openshift/v4_22: roll back to Ignition spec 3.5.0
   
You can look at the 4.19.0 rollback (https://github.com/coreos/butane/commit/ca25c01) or the stabilization commits on main for the naming style.
 
2. There's a doc issue with storage.trees -- see inline comment on butane.yaml.
Inline comment on internal/doc/butane.yaml, line 42 (the min: 4.23.0 change):
This change is correct (bumping the setuid "not supported" transform to start at 4.23), but I think there's a related issue with the trees description transform further down.
 
 At line 286, the transform that strips the file_mode/dir_mode prose from the trees description has max: 4.20.0:
 
>      - variant: openshift
>       max: 4.20.0
> 
 Since v4_22 (and v4_21) now both use base/v0_6 which doesn't have file_mode, dir_mode, user, or group on Tree, this should be max: 4.22.0 so the generated docs don't advertise fields that don't exist in the schema.
 
You can verify by looking at docs/config-openshift-v4_22.md line 140 -- it still says "Ownership, file modes (using file_mode) and directories modes (using dir_mode) can be specified for the tree" even though those fields were removed from the field listing. This looks like it was actually a pre-existing bug from v4_21 as well, but we should fix it here since this PR is what makes v4_22 affected.

yasminvalim:
@prestist  thanks for your review! I squashed the changes in one commit and fixed the message following the repo convention. I also updated the max version and the docs. Feel free to take a look on that :)

prestist (approved):
LGTM, thank you @yasminvalim! Commit message looks good and the trees doc fix for v4_21 is a nice bonus

docs/config-openshift-v4_21.md
```diff
@@ -137,7 +137,7 @@ The OpenShift configuration is a YAML document conforming to the following speci
         * **_needs_network_** (boolean): whether or not the device requires networking.
     * **_cex_** (object): describes the IBM Crypto Express (CEX) card configuration for the luks device.
       * **_enabled_** (boolean): whether or not to enable cex compatibility for luks. If omitted, defaults to false.
-  * **_trees_** (list of objects): a list of local directory trees to be embedded in the config. Symlinks must not be present. Ownership, file modes (using `file_mode`) and directories modes (using `dir_mode`) can be specified for the tree. If not specified, ownership is not preserved and file modes are set to 0755 if the local file is executable or 0644 otherwise. File attributes can be overridden by creating a corresponding entry in the `files` section; such entries must omit `contents`.
+  * **_trees_** (list of objects): a list of local directory trees to be embedded in the config. Symlinks must not be present. Ownership is not preserved. File modes are set to 0755 if the local file is executable or 0644 otherwise. File attributes can be overridden by creating a corresponding entry in the `files` section; such entries must omit `contents`.
```
travier, line 140: Why do we need this change here? This should have been in a distinct commit with an explaination.

docs/release-notes.md
```diff
@@ -14,6 +14,7 @@ nav_order: 9
 
 ### Misc. changes
 
+- Roll back to Ignition spec 3.5.0 _(openshift 4.22.0)_
```
travier, line 17: This is not a misc change, this would be a breaking change.

internal/doc/butane.yaml
```diff
@@ -284,7 +284,7 @@ root:
                 - variant: flatcar
                   max: 1.1.0
                 - variant: openshift
-                  max: 4.20.0
+                  max: 4.22.0
```
travier, line 287: Hum, that's a two version jump, so is this related to the docs change that we missed? This should have been in its own commit.

travier:
We should not squash unrelated changes into a single commit. We need the split into individual commits to understand what each change mean.

travier:
This should reference:
- The stabilization PR: We stabilized on this version to get support for 3.6.0 for a reason: https://github.com/coreos/butane/pull/704
- Why we need to undo that. What's the MCO issue? (Edit: https://redhat.atlassian.net/browse/OCPBUGS-90256)

travier:
This needs an entry in the upgrading docs to note the breaking change.

yasminvalim:
@travier can I revert this and create a new PR with those changes?

travier:
Yes please, make a PR that reverts this and re-splits it with fixes.

yasminvalim:
Revert PR: https://github.com/coreos/butane/pull/718

PR #718, opened 2026-06-26
Revert "Merge pull request #717 from coreos/update-4.22"

yasminvalim:
## Summary

Re-applies the OpenShift 4.22.0 rollback from #717, split into separate commits per review feedback on that PR.
OCP 4.22's Machine Config Operator does not support Ignition 3.6.0 ([OCPBUGS-90256](https://redhat.atlassian.net/browse/OCPBUGS-90256)), so `openshift` 4.22.0 is rolled back to base/v0_6 (Ignition 3.5.0).

Changes compared to #717:
- Split into 4 commits (rollback, trees doc fix, setuid transform, breaking-change docs)
- Document the rollback under **Breaking changes** instead of Misc
- Add upgrading guidance in `docs/upgrading-openshift.md`

---

Revert: Merge pull request #717 from coreos/update-4.22

This reverts commit 10436b9e0d79560511a411b417816cd7e58cd934, reversing changes made to 5d1a6d7.

---

docs: Permissions for `storage.trees` not available in openshift 4.21

Extended permissions for `storage.trees` were added in
https://github.com/coreos/butane/pull/644 to the base 0.7-exp spec.

The openshift 4.21 spec was stabilized on the fcos 1.6 & base 0.6 spec,
thus it did not include support for those. Unfortunately, we missed that
in https://github.com/coreos/butane/pull/674.

---

openshift/v4_22: Roll back to Ignition spec 3.5.0

OpenShift 4.22's Machine Config Operator does not fully support Ignition
3.6.0 yet (OCPBUGS-90256).

Move v4_22 back to base/v0_6 and fcos/v1_6 and ignition/v3_5.

We are doing this breaking change in a minor release as the stabilized
spec that was released was unfortunately not working and thus there
should be no configuration from this spec in use so far.

What was moved back:
- Permissions for `storage.trees` not available in openshift 4.22
  (depends on base/v0_7 & fcos/v1_7, see:
  https://github.com/coreos/butane/pull/644)
- Remove support for special mode bits (ignition/v3_6, see
  https://coreos.github.io/ignition/migrating-configs/#from-version-350-to-360)

Fixes: https://github.com/coreos/butane/pull/704
Fixes: https://redhat.atlassian.net/browse/OCPBUGS-90256

---

internal/doc/butane: Add some notes about support

- Note dependency on base 0.7 for storage.trees permissions
- Note dependency on external MCO support for GRUB config

See: https://github.com/coreos/butane/issues/719

### DIFF
--- a/config/openshift/v4_22/result/schema.go
+++ b/config/openshift/v4_22/result/schema.go
@@ -15,7 +15,7 @@
 package result
 
 import (
-	"github.com/coreos/ignition/v2/config/v3_6/types"
+	"github.com/coreos/ignition/v2/config/v3_5/types"
 )
 
 const (
--- a/config/openshift/v4_22/result/schema.go
+++ b/config/openshift/v4_22/result/schema.go
@@ -15,7 +15,7 @@
 package result
 
 import (
-	"github.com/coreos/ignition/v2/config/v3_5/types"
+	"github.com/coreos/ignition/v2/config/v3_6/types"
 )
 
 const (
--- a/config/openshift/v4_22/schema.go
+++ b/config/openshift/v4_22/schema.go
@@ -15,7 +15,7 @@
 package v4_22
 
 import (
-	fcos "github.com/coreos/butane/config/fcos/v1_7"
+	fcos "github.com/coreos/butane/config/fcos/v1_6"
 )
 
 const ROLE_LABEL_KEY = "machineconfiguration.openshift.io/role"
--- a/config/openshift/v4_22/schema.go
+++ b/config/openshift/v4_22/schema.go
@@ -15,7 +15,7 @@
 package v4_22
 
 import (
-	fcos "github.com/coreos/butane/config/fcos/v1_6"
+	fcos "github.com/coreos/butane/config/fcos/v1_7"
 )
 
 const ROLE_LABEL_KEY = "machineconfiguration.openshift.io/role"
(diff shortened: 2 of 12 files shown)


travier:
Let's not just revert but re-add the changes from https://github.com/coreos/butane/pull/717 in this PR as split commits.

yasminvalim:
@travier @prestist CI fails on ./generate because the manual storage.trees doc fix gets reverted while butane.yaml still has max: 4.20.0. It only passes with max: 4.22.0. Should we bump the transform to 4.22.0, or keep the manual doc edits?

travier:
> undoing the stabilization in #704.

So it's not exactly undoing the stabilization but stabilizing on another version instead.

travier:
> openshift/v4_22: roll back to Ignition spec 3.5.0

For this commit message, let's explicitly link to the full URL for OCPBUGS-90256:

> See: https://redhat.atlassian.net/browse/OCPBUGS-90256

Same note about "undoing" as well.

We may squash the commit "docs: note openshift 4.22.0 Ignition rollback breaking change" with this one as this is a single logical change (the docs changes are part of the code change).

---

> docs: fix storage.trees description for openshift 4.21–4.22

This should be included in the commit that adds the related code change as the docs are automatically generated.

This should reference the PRs/commits where we missed that change.

---

So I would do this in a different order:
- Revert (OK as is)
- First, fix what's incorrect in 4.21 only and explain what we missed and why. Can we figure out a way to not miss it in the future?
- Then do the 4.22 rollback to 3.5.0. It might be a single commit here in the end that would include notes about all the changes in its description.

travier:
I've created https://github.com/coreos/butane/issues/719 to track what is not supported by the MCO and must be removed on each stabilization from the openshift spec.

prestist (approved):
Thank you both for working on this. LGTM.

PR #722, opened 2026-07-09
Correct partition index loop in translate.go

skewballfox:
currently trying to override the default sizes for the boot partitions. 

In my butain file, I have:
```
        - number: 1
          label: bios
          size_mib: 1
          type_guid: 21686148-6449-6E6F-744E-656564454649
        - number: 2
          label: efi 
          size_mib: 1024
          type_guid: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
          wipe_partition_entry: true
        - number: 3
          label: boot
          size_mib: 2048
          type_guid: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
          wipe_partition_entry: true
        - label: root
          number: 4
          wipe_partition_entry: true
          type_guid: 4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709
```
which should be valid. root is the last partition, and all the others have a set size. currently this throws:
```
     |                  ^^^^ root partition cannot expand; it is set to fill available space but is followed by an auto-positioned partition
 122 |           number: 4
 123 |           start_mib: 0
     |
Config produced warnings and --strict was specified
```
this shoud only check the partitions after root

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined root-partition constraint validation during configuration conversion by limiting the check to partitions after the `root` entry, preventing incorrect or overly broad warnings for valid disk layouts.
  * Added/extended test coverage to confirm that when `root` is the last partition (and preceding partitions have explicit, non-zero sizes), no root constraint warning is emitted.
* **Documentation**
  * Updated the release notes to reflect the root-partition constraint fix.

### DIFF
--- a/config/fcos/v1_3/translate.go
+++ b/config/fcos/v1_3/translate.go
@@ -75,10 +75,7 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 			if partition.Label != nil {
 				if *partition.Label == "root" {
 					if partition.SizeMiB == nil || *partition.SizeMiB == 0 {
-						for idx := range disk.Partitions {
-							if idx == p {
-								continue
-							}
+						for idx := p + 1; idx < len(disk.Partitions); idx++ {
 							if disk.Partitions[idx].StartMiB == nil || *disk.Partitions[idx].StartMiB == 0 {
 								r.AddOnWarn(path.New("json", "storage", "disks", i, "partitions", p, "label"), common.ErrRootConstrained)
 								break
--- a/config/fcos/v1_3/translate_test.go
+++ b/config/fcos/v1_3/translate_test.go
@@ -1652,6 +1652,37 @@ func TestRootPartitionConstraints(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "root last with sized partitions before",
+			in: Config{
+				Config: base.Config{
+					Storage: base.Storage{
+						Disks: []base.Disk{
+							{
+								Device: "/dev/vda",
+								Partitions: []base.Partition{
+									{
+										Label:   util.StrToPtr("bios"),
+										Number:  1,
+										SizeMiB: util.IntToPtr(1),
+									},
+									{
+										Label:   util.StrToPtr("boot"),
+										Number:  3,
+										SizeMiB: util.IntToPtr(2048),
+									},
+									{
+										Label:  util.StrToPtr("root"),
+										Number: 4,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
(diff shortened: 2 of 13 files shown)


skewballfox:
btw, I tried compiling locally and testing out this with the butane file containing the code above, for some reason even though this should start from the partitions after root (if any), it seems the same error message, so I'm guessing either the array of partitions isn't in the declared order, or some partition is part of that array that isn't one the ones defined.

prestist:
Thanks for the fix @skewballfox! I agree this is a bug; the loop should only check partitions declared after root, since sgdisk processes them in declaration order and only subsequent auto-positioned partitions would constrain root's expansion. This won't break anyone since it only removes a false warning.
A couple things needed before we can merge:
1. Apply the fix to all spec versions
The same bug exists in `v1_3`, `v1_4`, `v1_6`, `v1_7`, and `v1_8_exp` (config/fcos/*/translate.go). Since this is a bug fix and won't change behavior for configs that are currently valid, it should be safe to fix everywhere.
2. Add a test case
Could you add a test case to TestRootPartitionConstraints where root is the last partition with sized partitions before it (matching your original scenario)? This would prevent the regression. Something like:
```
{
	name: "root last with sized partitions before",
	in: Config{
		Config: base.Config{
			Storage: base.Storage{
				Disks: []base.Disk{
					{
						Device: "/dev/vda",
						Partitions: []base.Partition{
							{
								Label:   util.StrToPtr("bios"),
								Number:  1,
								SizeMiB: util.IntToPtr(1),
							},
							{
								Label:   util.StrToPtr("boot"),
								Number:  3,
								SizeMiB: util.IntToPtr(2048),
							},
							{
								Label:  util.StrToPtr("root"),
								Number: 4,
							},
						},
					},
				},
			},
		},
	},
	report: report.Report{}, // no warning expected
},
```
3. **Regarding the warning still appearing**
You mentioned the fix didn't resolve the warning when you tested. Since this PR only patches v1_5, if your Butane config uses a different version (e.g., version: "1.6.0"), it would hit the unfixed code. Fixing all versions (point 1) should resolve that.
4. Commit conventions
The project uses the format component: description -- for this change: */translate: fix root partition constraint check to only examine subsequent partitions

skewballfox:
sorry, been afk a bit. I'll push the changes today or tomorrow

angelcerveraroldan:
@skewballfox The `butane` repository has been moved into [`ignition/butane`](https://github.com/coreos/ignition/tree/main/butane). Would you like to move this PR to that repo ? I'd be happy to continue the review there.

skewballfox:
patched it over, I'll go ahead and close it here

PR #725, opened 2026-08-01
base: validate local ignition config files under ignition.config

### LINKED ISSUE #724: bug: Dead validation code: local ignition config files never validated
---
---
name: release checklist
about: release checklist template
title: Dead validation code: local ignition config files never validated
labels: jira,kind/bug
warning: |
    ⚠️ Template generated by https://github.com/coreos/repo-templates; do not edit downstream
---

## Bug Report

**File:** `base/v0_8_exp/translate.go`, line 157 (same pattern at `base/v0_7/translate.go:153`)

**Summary:**

`translateResource` contains dead validation code. The `strings.HasPrefix` guard at line 157 is always `false`, so `ValidateIgnitionConfig` is never called for local files — even when those files are used in `ignition.config.merge` or `ignition.config.replace` contexts. Invalid or corrupt ignition config data in a `local` field passes through translation silently with no warning or error.

---

### Buggy code

https://github.com/coreos/butane/blob/cb34e120e5267bfd5bdfa83fa8c2e44e06dedda2/base/v0_8_exp/translate.go#L157

```go
if from.Local != nil {
    c := path.New("yaml", "local")
    contents, err := baseutil.ReadLocalFile(*from.Local, options.FilesDir)
    if err != nil {
        r.AddOnError(c, err)
        return
    }
    // Validating the contents of the local file from here since there is no way to
    // get both the filename and filedirectory in the Validate context
    if strings.HasPrefix(c.String(), "$.ignition.config") {   // ← always false
        rp, err := ValidateIgnitionConfig(c, contents)
        r.Merge(rp)
        if err != nil {
            return
        }
    }
    ...
}
```

---

### Root cause

`c` is constructed as `path.New("yaml", "local")`. The first argument is the **tag**; the second is a **path element**.

`path.ContextPath.String()` (defined in `vendor/github.com/coreos/vcontext/path/path.go:36`) builds its output by joining path **elements** with dots, prefixed with `$`. The tag is **not** included in the string representation.

Therefore:

```
path.New("yaml", "local").String()  →  "$.local"
strings.HasPrefix("$.local", "$.ignition.config")  →  false
```

The condition can never be `true` regardless of where `translateResource` is called from. `ValidateIgnitionConfig` is dead code for all `local` field inputs.

---

### Impact

Local files referenced via the `local` field inside `ignition.config.merge` or `ignition.config.replace` blocks are never validated as Ignition configs at translation time. Malformed or invalid ignition config JSON/YAML in those files is silently accepted and encoded into the output without any diagnostic.

---

### Reproduction

Confirmed by evaluating the vendored vcontext library directly:

```go
import "github.com/coreos/vcontext/path"
import "strings"

c := path.New("yaml", "local")
fmt.Println(c.String())                                         // "$.local"
fmt.Println(strings.HasPrefix(c.String(), "$.ignition.config")) // false
```

`path.New("yaml", "local").String()` returns `"$.local"`, which does not satisfy `strings.HasPrefix(..., "$.ignition.config")`.

The same dead-code pattern exists independently at `base/v0_7/translate.go:153`.

---

Tagging:
 - [ ] Confirm `path.ContextPath.String()` does not include the tag in any vcontext version vendored by this repo
 - [ ] Confirm `ValidateIgnitionConfig` is unreachable for `local` field inputs in both `base/v0_7` and `base/v0_8_exp`
 - [ ] Confirm no test currently exercises the `ValidateIgnitionConfig` branch via a `local` field

Fedora packaging:
 - [ ] No response

GitHub release:
 - [ ] No response

Quay release:
 - [ ] No response

deepak0x:
## Summary

`translateResource` contained dead validation code: the `strings.HasPrefix(c.String(), "$.ignition.config")` guard at `base/v0_8_exp/translate.go:157` (and `base/v0_7/translate.go:153`) is always `false`, because `c` is `path.New("yaml", "local")` whose `String()` is `$.local`, never `$.ignition.config`. As a result, local files referenced via `ignition.config.merge` / `ignition.config.replace` were never validated as Ignition configs, and malformed configs passed through silently (issue #724).

## Fix

Refactor `translateResource` into an internal helper `translateResourceInternal` with a `validateIgnitionConfig bool` parameter. Register `translateResourceForIgnition` (which passes `true`) for the `ignition.config` translation path, while the existing `translateResource` (passing `false`) continues to serve other local resources such as file contents, where the content is not an Ignition config and must not be validated as one.

The local file is now validated via `ValidateIgnitionConfig` during translation, where `options.FilesDir` is available (validation has no access to `FilesDir`, which is why the check belongs in the translation phase).

## Test plan

- Added `TestTranslateIgnitionLocalValidation` to both `base/v0_7` and `base/v0_8_exp` translate tests, covering:
  - malformed local config under `merge` → rejected
  - valid local config under `merge` → accepted
  - malformed local config under `replace` → rejected

All existing tests in `base/v0_7` and `base/v0_8_exp` continue to pass.

Fixes: #724

## Summary by CodeRabbit

* **Bug Fixes**
  * Local Ignition configuration files referenced by `merge` and `replace` are now validated as Ignition configs.
  * Malformed nested configurations produce translation errors, while valid configurations continue to translate successfully.

* **Tests**
  * Added coverage for valid and invalid local Ignition configuration scenarios.

### DIFF
--- a/base/v0_7/translate.go
+++ b/base/v0_7/translate.go
@@ -112,7 +112,7 @@ func (c Config) ToIgn3_6Unvalidated(options common.TranslateOptions) (types.Conf
 
 func translateIgnition(from Ignition, options common.TranslateOptions) (to types.Ignition, tm translate.TranslationSet, r report.Report) {
 	tr := translate.NewTranslator("yaml", "json", options)
-	tr.AddCustomTranslator(translateResource)
+	tr.AddCustomTranslator(translateResourceForIgnition)
 	to.Version = types.MaxVersion.String()
 	tm, r = translate.Prefixed(tr, "config", &from.Config, &to.Config)
 	translate.MergeP(tr, tm, &r, "proxy", &from.Proxy, &to.Proxy)
@@ -135,6 +135,18 @@ func translateFile(from File, options common.TranslateOptions) (to types.File, t
 }
 
 func translateResource(from Resource, options common.TranslateOptions) (to types.Resource, tm translate.TranslationSet, r report.Report) {
+	return translateResourceInternal(from, options, false)
+}
+
+// translateResourceForIgnition is like translateResource but additionally
+// validates the contents of local files as Ignition configs, since the
+// resources under ignition.config.merge/replace are themselves Ignition
+// configs.
+func translateResourceForIgnition(from Resource, options common.TranslateOptions) (to types.Resource, tm translate.TranslationSet, r report.Report) {
+	return translateResourceInternal(from, options, true)
+}
+
+func translateResourceInternal(from Resource, options common.TranslateOptions, validateIgnitionConfig bool) (to types.Resource, tm translate.TranslationSet, r report.Report) {
 	tr := translate.NewTranslator("yaml", "json", options)
 	tm, r = translate.Prefixed(tr, "verification", &from.Verification, &to.Verification)
 	translate.MergeP2(tr, tm, &r, "http_headers", &from.HTTPHeaders, "httpHeaders", &to.HTTPHeaders)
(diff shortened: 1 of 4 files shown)


yasminvalim:
Hey, @deepak0x! Thanks for the PR! 
We’re archiving this repository as Butane is being merged into Ignition, so new changes should now go to Ignition:
https://github.com/coreos/ignition

You’re very welcome to reopen this work there if it makes sense. I will be glad to review! Thanks again for contributing! :cat:

